PR #20, opened 2021-10-20
support output stack

duanjiong:
1. add options --output-stack
2. Added a bpf map of type BPF_MAP_TYPE_STACK_TRACE to hold the stack, with a maximum stack depth of 50
3. sort the symbols in /proc/kallsyms, then find the closest symbol and print it

### DIFF
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -272,7 +272,7 @@ set_output(struct pt_regs *ctx, struct sk_buff *skb, struct event_t *event, stru
 		set_skb_btf(skb, &event->print_skb_id);
 
 	if (cfg->output_stack) {
-		event->print_stack_id = bpf_get_stackid(ctx, &print_stack_map, BPF_F_REUSE_STACKID | BPF_F_FAST_STACK_CMP);
+		event->print_stack_id = bpf_get_stackid(ctx, &print_stack_map, BPF_F_FAST_STACK_CMP);
 	}
 }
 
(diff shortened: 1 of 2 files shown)


bpf/kprobe_pwru.c
```diff
@@ -76,10 +76,18 @@ struct config {
...
-
+	u8 output_stack;
+	u8 pad;
 } __attribute__((packed));
 
+#define MAX_STACK_DEPTH 50
+struct {
+	__uint(type, BPF_MAP_TYPE_STACK_TRACE);
+	__uint(max_entries, 256);
+	__uint(key_size, sizeof(u32));
```
brb, line 87: Nit: you can replace with `__type(key, u32));`. Same goes for the val.

bpf/kprobe_pwru.c
```diff
@@ -262,6 +270,10 @@ set_output(struct sk_buff *skb, struct event_t *event, struct config *cfg) {
 
 	if (cfg->output_skb)
 		set_skb_btf(skb, &event->print_skb_id);
+
+	if (cfg->output_stack) {
+		event->print_stack_id = bpf_get_stackid(ctx, &print_stack_map, BPF_F_REUSE_STACKID | BPF_F_FAST_STACK_CMP);
```
brb, line 275: I'd avoid setting `BPF_F_REUSE_STACKID`, as otherwise we risk to print unrelated stacks.

duanjiong: done

PR #22, opened 2021-10-21
add option --filter-func

duanjiong:
It is a regular expression that is used to filter the probe function

### DIFF
--- a/internal/pwru/types.go
+++ b/internal/pwru/types.go
@@ -12,6 +12,7 @@ const (
 
 type Flags struct {
 	FilterMark    uint32
+	FilterFunc    string
 	FilterProto   string
 	FilterSrcIP   string
 	FilterDstIP   string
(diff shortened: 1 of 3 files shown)


internal/pwru/utils.go
```diff
@@ -20,6 +21,14 @@ func GetFuncs() (Funcs, error) {
 
 	callback := func(typ btf.Type) {
 		fn := typ.(*btf.Func)
+
+		if pattern != "" {
+			match, _ := regexp.Match(pattern, []byte(fn.Name))
```
brb, line 26: I think we can optimize it by compiling the regexp first (with `regexp.Compile`), and then matching on the compiled regexp.

internal/pwru/types.go
```diff
@@ -26,6 +27,7 @@ type Flags struct {
 }
 
 func (f *Flags) SetFlags() {
+	flag.StringVar(&f.FilterFunc, "filter-func", "", "filter the kernel functions that can be probed, and it is a regular expression")
```
brb, line 30:
```suggestion
	flag.StringVar(&f.FilterFunc, "filter-func", "", "filter the kernel functions that can be probed; the filter can be a regular expression (RE2)")
```

duanjiong: done

PR #26, opened 2021-10-21
support filter by ip6 address

duanjiong:
1. ipv6 is a bit tricky to parse out the transport layer protocols, so we don't support transport layer filtering for now.
2. rename filter_mark to filter_meta, other metadata should be supported later


brb:
@duanjiong Hi, could you ACK https://github.com/cilium/pwru/pull/190?

Issue #23, opened 2021-10-21
support filter by net namespace


duanjiong:
Network addresses may overlap between multiple net namespaces, which can interfere with our filtering.

Asphaltt:
> assign netns ID to an existing netns to facilitate your proposed debugging?

By looking down the kernel source code [net_namespace/peernet2id](https://elixir.bootlin.com/linux/v5.15.3/source/net/core/net_namespace.c#L267), **netnsid** is not retrieved from `struct net` neither `struct ns_common`.

**inode** is a good way to identify an net namespace. In kernel space, inode is the `unsigned int inum` field of `struct ns_common`. In user space, inode is the inode of the net namespace file, wherever the file is, like `/proc/<pid>/ns/net`, `/var/run/docker/netns/<netns name>`.

For example:

log of pwru: *0xffff8f26c23d4500 [isc-worker0000]     neigh_resolve_output    3355665098061 netns=4026531992 mark=0x0 ifindex=2 proto=8 mtu=1500 len=62*

inode of netns file: *got netns: NS(4:4026531992)* 

(Go code to get inode: [ns.UniqueID](https://github.com/vishvananda/netns/blob/50045581ed74/netns.go#L52))

brb:
@Asphaltt Thanks! Using the inode makes sense to me. Would you like to work on this?

brb:
Fixed by #34.

PR #34, opened 2021-11-20
support output netns inode

Asphaltt:
#23

### DIFF
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -32,12 +32,13 @@ union addr {
 } __attribute__((packed));
 
 struct skb_meta {
+	u32 netns;
 	u32 mark;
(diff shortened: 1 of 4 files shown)


brb (changes requested):
Thanks! Could you sign-off your commit messages (`git commit -s`)?

Asphaltt:
> Could you sign-off your commit messages (git commit -s)?

Sorry, I got it.

brb (approved):
Thanks again! :rocket: 

For future users, don't mix inode number of a symlink vs the actual netns number. I.e., `ls -li /proc/$PID/ns/net` returns the symlink's inode. To get the actual inode, you can use `ls -al`, and then take the number from the `net -> net[$INODE]` output.

PR #36, opened 2021-11-24
add option --output-limit

Asphaltt:
1. Limit the number of printing to prevent running program from unstopping.
   The default number is 1000. The number 0 is for nearly unlimited printing.
2. update README.md

Signed-off-by: Leon Huayra <hffilwlqm@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,9 +98,11 @@ Usage of ./pwru:
       --filter-dst-port uint16      filter destination port
       --filter-func string          filter the kernel functions that can be probed; the filter can be a regular expression (RE2)
       --filter-mark uint32          filter skb mark
+      --filter-netns uint32         filter netns inode
       --filter-proto string         filter L4 protocol (tcp, udp, icmp)
       --filter-src-ip string        filter source IP addr
       --filter-src-port uint16      filter source port
+      --output-limit-lines uint     exit the program after the number of events has been received/printed
       --output-meta                 print skb metadata
       --output-relative-timestamp   print relative timestamp per skb
       --output-skb                  print skb
--- a/main.go
+++ b/main.go
@@ -51,10 +51,6 @@ func main() {
 	}
 
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
-	go func() {
-		<-ctx.Done()
-		log.Println("Received signal, exiting program..")
-	}()
 	defer stop()
 
 	funcs, err := pwru.GetFuncs(flags.FilterFunc)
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -98,9 +98,11 @@ Usage of ./pwru:
       --filter-dst-port uint16      filter destination port
       --filter-func string          filter the kernel functions that can be probed; the filter can be a regular expression (RE2)
       --filter-mark uint32          filter skb mark
+      --filter-netns uint32         filter netns inode
       --filter-proto string         filter L4 protocol (tcp, udp, icmp)
       --filter-src-ip string        filter source IP addr
       --filter-src-port uint16      filter source port
+      --output-limit uint           limit the number of printing, 0 for unlimited (default 1000)
```
brb, line 105: - Let's make the default 0. Otherwise, the existing users might not notice why it stopped after 1k lines.
- Maybe `--output-stop-after-num-lines` is more clear name?

main.go
```diff
@@ -162,8 +158,24 @@ func main() {
 	output := pwru.NewOutput(&flags, printSkbMap, printStackMap, addr2name)
 	output.PrintHeader()
 
+	outputLimit := flags.OutputLimit
+	if outputLimit == 0 {
+		outputLimit--
```
brb, line 163: Why this is needed?

brb (changes requested):
Thanks! A few comments.

main.go
benner: From code below:
```
for ; outputLimit != 0; outputLimit-- {
```
Negative `outputLimit` decrement will never (unless overflow) reach zero.

README.md
Asphaltt: >  Maybe --output-stop-after-num-lines is more clear name?

parameter name should be short and clear. `--output-limit-lines` better?

main.go
Asphaltt: > Why this is needed?

when `outputLimit` is 0, convert `outputLimit` to 1<<64 - 1. check [https://go.dev/play/p/Cp8Anfq8xBv](https://go.dev/play/p/Cp8Anfq8xBv)

So on, the following `for` loop seems like an infinite loop.


> Negative outputLimit decrement will never (unless overflow) reach zero.

`outputLimit` is an `uint64`, which won't be negative.

brb: > loop seems like an infinite loop.

In theory it won't be infinite. `outputLimit = uint64(0); outputLimit--` makes it `max(uint64))`. I'd prefer to avoid such hacks.

README.md
brb: `--output-limit-lines` is better. Could you change the default (to 0), and update the description of the flag to `Exit the program after the number of events has been received/printed`?

main.go
```diff
@@ -51,10 +52,6 @@ func main() {
 	}
 
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
-	go func() {
```
brb, line 54: Let's keep this here.

main.go
```diff
@@ -162,8 +159,24 @@ func main() {
 	output := pwru.NewOutput(&flags, printSkbMap, printStackMap, addr2name)
 	output.PrintHeader()
 
+	outputLimitLines := flags.OutputLimitLines
```
brb, line 162: Sorry, I was not clear enough. But let's do something like:
```
exitAfterNumLines := flags.OutputLimitLines != 0

[...]

for i := 0; i < flags.OutputLimitLines || !exitAfterLines; i++ {
```

main.go
Asphaltt: After printing `--output-limit-lines` lines, "Received signal, exiting program.." should not be printed.

So while exiting, "Received signal, exiting program.." or "Printed %d events, exiting program.." is printed. They shouldn't be printed both.

brb (approved):

PR #40, opened 2021-12-14
bpf/headers: add script to update headers from libbpf

tklauser:
Add an update.sh script which can be used to update the libbpf headers
used by pwru. Pin it to libbpf 0.5.0 which seems to be the closest
matching version. Updating to that version leads to some benign changes
to bpf_helper_defs.h.

This is based on the script in github.com/cilium/ebpf.


brb (approved):
Thanks!

PR #43, opened 2021-12-17
ARM64 support

tklauser:
Allow running `pwru` on ARM64 platforms.

Verified on a Raspberry Pi with kernel 5.10. I still observe some issues with events seemingly reported for unrelated processes, probably an issue with the PID->process lookup.

In any case, it's a start. See commits for further details.


brb (approved):
:rocket:

PR #44, opened 2021-12-22
internal/pwru: Match entire string for --filter-func

ypl-coffee:
Hi all,

Currently `--filter-func=kfree_skb` matches all functions whose name
contains the `"kfree_skb"` string, including `__kfree_skb()`, `kfree_skbmem()`
etc.  To match `kfree_skb()` only, one has to do something like
`--filter-func="^kfree_skb$"`.

This is counterintuitive.  Match the entire string instead in
`GetFuncs()`, so that `--filter-func=kfree_skb` only matches `kfree_skb()`.
To match all function names containing `"kfree_skb"`, one should do
something like `--filter-func=".*kfree_skb.*"` instead.

Thanks,
Signed-off-by: Peilin Ye <peilin.ye@bytedance.com>

### DIFF
--- a/internal/pwru/utils.go
+++ b/internal/pwru/utils.go
@@ -26,8 +26,9 @@ func GetFuncs(pattern string) (Funcs, error) {
 	}
 	callback := func(typ btf.Type) {
 		fn := typ.(*btf.Func)
+		fnName := string(fn.Name)
 
-		if pattern != "" && !reg.Match([]byte(fn.Name)) {
+		if pattern != "" && reg.FindString(fnName) != fnName {
 			return
 		}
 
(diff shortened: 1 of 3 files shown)


brb (changes requested):
Hey, thanks for the PR and sorry for the delay.

I'm fine with the change. Could you update the flag description and README.md to mention the exact match and how wildcarded match can be set?

internal/pwru/utils.go
```diff
@@ -26,8 +26,9 @@ func GetFuncs(pattern string) (Funcs, error) {
 	}
 	callback := func(typ btf.Type) {
 		fn := typ.(*btf.Func)
+		fnName := string(fn.Name)
 
-		if pattern != "" && !reg.Match([]byte(fn.Name)) {
+		if pattern != "" && reg.FindString(fnName) != fnName {
```
gandro, line 31: `FindString` does a prefix match. Is that intended? With that code, `--filter-func=kfree_sbk would still match kfree_skbmem` https://go.dev/play/p/M3yIKZpCVf8

ypl-coffee:
> Hey, thanks for the PR and sorry for the delay.

Thanks for the review.

> I'm fine with the change. Could you update the flag description and README.md to mention the exact match and how wildcarded match can be set?

Sure of course, I just did a force-push.

brb (approved):
Thanks!

PR #48, opened 2022-01-24
Makefile: fix release target

tklauser:
Run `go generate` and install clang in the container to make the release
target actually work. Also remove generated files as part of the clean
target.

Fixes: ce9cb04ac6ac ("Add minimal Makefile")


brb (approved):
Thanks!

Issue #46, opened 2022-01-21
Perf event ring buffer full, dropped 16 samples 


BurlyLuo:
```
root@bpf2:~# ./pwru --filter-proto icmp  --output-tuple
2022/01/21 04:44:15 Attaching kprobes...
577 / 577 [---------------------------------------------------------------------------------------------------------------------------------] 100.00% 34 p/s
2022/01/21 04:44:32 Attached (ignored 13)
2022/01/21 04:44:32 Listening for events..
               SKB         PROCESS                     FUNC        TIMESTAMP
0xffff96f353734300          [ping]           __ip_local_out   52052971636335 10.0.1.30:0->10.0.1.23:0(icmp)
0xffff96f353734300          [ping]                ip_output   52052971655120 10.0.1.30:0->10.0.1.23:0(icmp)
0xffff96f353734300          [ping]             nf_hook_slow   52052971658827 10.0.1.30:0->10.0.1.23:0(icmp)
0xffff96f353734300          [ping] __cgroup_bpf_run_filter_skb   52052971662013 10.0.1.30:0->10.0.1.23:0(icmp)
0xffff96f353734300          [ping]     neigh_resolve_output   52052971666902 10.0.1.30:0->10.0.1.23:0(icmp)
0xffff96f353734300          [ping]       __neigh_event_send   52052971669217 10.0.1.30:0->10.0.1.23:0(icmp)
0xffff96f353734300          [ping]               eth_header   52052971673094 10.0.1.30:0->10.0.1.23:0(icmp)
0xffff96f353734300          [ping]                 skb_push   52052971676681 10.0.1.30:0->10.0.1.23:0(icmp)
0xffff96f353734300          [ping]           dev_queue_xmit   52052971678785 10.0.1.30:0->10.0.1.23:0(icmp)
0xffff96f353734300          [ping]      netdev_core_pick_tx   52052971681881 10.0.1.30:0->10.0.1.23:0(icmp)
0xffff96f353734300          [ping]       netif_skb_features   52052971684686 10.0.1.30:0->10.0.1.23:0(icmp)
0xffff96f353734300          [ping]  passthru_features_check   52052971686700 10.0.1.30:0->10.0.1.23:0(icmp)
0xffff96f353734300          [ping]     skb_network_protocol   52052971689855 10.0.1.30:0->10.0.1.23:0(icmp)
[...]

PR #55, opened 2022-03-12
Add argument per-cpu-buffer and fix Vagrant setup

dezmodue:
Following up on this comment https://github.com/cilium/pwru/issues/46#issuecomment-1023213310 adding a new command line argument to set the per CPU buffer. Fixed an issue in the Vagrant setup where llvm-strip is not in $PATH and make fails.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -8,5 +8,7 @@ Vagrant.configure("2") do |config|
       apt-get update
       apt-get install -y clang-12 golang make
       update-alternatives --install /usr/bin/clang clang /usr/bin/clang-12 100
+      [ -f /usr/lib/llvm-12/bin/llvm-strip ] && [ -f /usr/local/bin/llvm-strip ] || \
+      ln -s /usr/lib/llvm-12/bin/llvm-strip /usr/local/bin/llvm-strip
     SHELL
 end
--- a/internal/pwru/types.go
+++ b/internal/pwru/types.go
@@ -30,6 +30,8 @@ type Flags struct {
 	OutputSkb        bool
 	OutputStack      bool
 	OutputLimitLines uint64
+
+	PerCPUBuffer	uint32
 }
 
 func (f *Flags) SetFlags() {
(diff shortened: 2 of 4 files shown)


aroulin:
Can we fix the docker build also? The llvm-strip issue is there as well. I can send a separate PR if needed.

Also should we increase buffer size to larger than 4096 (page size)? Default page size for me shows a lot of dropped samples on a simple received ICMP packet.

internal/pwru/types.go
```diff
@@ -48,6 +50,7 @@ func (f *Flags) SetFlags() {
 	flag.BoolVar(&f.OutputSkb, "output-skb", false, "print skb")
 	flag.BoolVar(&f.OutputStack, "output-stack", false, "print stack")
 	flag.Uint64Var(&f.OutputLimitLines, "output-limit-lines", 0, "exit the program after the number of events has been received/printed")
+	flag.Uint32Var(&f.PerCPUBuffer, "per-cpu-buffer", 0, "per CPU buffer in bytes")
```
brb, line 53: Nit: instead of the default `0`, you could use `os.Getpagesize()`. This would also allow us to get rid of the if-statement at the line 104 in `.main.go`

Vagrantfile
```diff
@@ -8,5 +8,7 @@ Vagrant.configure("2") do |config|
       apt-get update
       apt-get install -y clang-12 golang make
       update-alternatives --install /usr/bin/clang clang /usr/bin/clang-12 100
+      [ -f /usr/lib/llvm-12/bin/llvm-strip ] && [ -f /usr/local/bin/llvm-strip ] || \
+      ln -s /usr/lib/llvm-12/bin/llvm-strip /usr/local/bin/llvm-strip
```
brb, line 12: Could you add into the commit message any info what this is fixing?

brb (changes requested):
Thanks! Some minor comments.

> Can we fix the docker build also? The llvm-strip issue is there as well. I can send a separate PR if needed.

@aroulin @dezmodue could you guys coordinate who is going to fix the issue? Thanks.

internal/pwru/types.go
dezmodue: Excellent point - will do

Vagrantfile
dezmodue: 👍

dezmodue:
I can take care of fixing also the Dockerfile @aroulin

dezmodue:
Updated PR - let me know if it addresses all the points raised

brb (approved):
Thanks!

aroulin:
Awesome @dezmodue thank you much

PR #69, opened 2022-06-24
Silence clang build warnings and fix Makefile release target

chendotjs:
1. Running either `make release` or `make local-release` will cause clang to emit warning messages of type `-Waddress-of-packed-member`. Silence the warnings by explicitly ignoring it.

2. Running `make release` will get error message like `Error: exec: "llvm-strip": executable file not found in $PATH`, fixed by installing `llvm`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ release:
 		--rm \
 		--workdir /pwru \
 		--volume `pwd`:/pwru docker.io/library/golang:1.18.2-alpine3.15 \
-		sh -c "apk add --no-cache make git clang && git config --global --add safe.directory /cilium && make local-release VERSION=${VERSION}"
+		sh -c "apk add --no-cache make git clang llvm && git config --global --add safe.directory /cilium && make local-release VERSION=${VERSION}"
 
 local-release: clean
 	OS=linux; \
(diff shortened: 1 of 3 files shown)


bpf/kprobe_pwru.c
```diff
@@ -348,7 +348,7 @@ set_output(struct pt_regs *ctx, struct sk_buff *skb, struct event_t *event, stru
 		set_tuple(skb, &event->tuple);
 
 	if (cfg->output_skb)
-		set_skb_btf(skb, &event->print_skb_id);
+		set_skb_btf(skb, (typeof(print_skb_id) *)(void *)&event->print_skb_id);
```
tklauser, line 351: I'd rather disable the warning explicitly using `-Wno-address-of-packed-member` to make it explicit we're silencing/ignoring it. With the current solution using the `void *` cast it's less obvious IMO.

tklauser (changes requested):

bpf/kprobe_pwru.c
chendotjs: @tklauser done as suggested.

tklauser (approved):
Thanks @chendotjs!

PR #71, opened 2022-06-30
Run with specified kernel BTF file

Asphaltt:
`pwru` can not runs in the Linux server, whose kernel is built without
`CONFIG_DEBUG_INFO_BTF=y`. But `pwru` can runs with specified kernel BTF
file in such Linux server.

I tried to run `pwru` on Ubuntu 20.04, whose kernel is 5.4 and is built
without `CONFIG_DEBUG_INFO_BTF=y`. Then, I downloaded the kernel BTF
file from github.com/aquasecurity/btfhub, and ran `pwru` with
`--kernel-btf`.

Unfortunately, I waited minutes to exit `pwru`. So I add a progress bar
when exit.

- add option `--kernel-btf`
- print a progress bar while detaching kprobes

Signed-off-by: Leon Hwang <hffilwlqm@gmail.com>

### DIFF
--- a/internal/pwru/types.go
+++ b/internal/pwru/types.go
@@ -5,8 +5,9 @@
 package pwru
 
 import (
-	flag "github.com/spf13/pflag"
 	"os"
+
+	flag "github.com/spf13/pflag"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


tklauser (approved):
LGTM, thank you.

Issue #78, opened 2022-08-10
Strip down pwru docker image


brb:
The Dockerfile ends up generating the 1.47GB image. This can be reduced by:

- Use `base` image instead of `ubuntu`.
- Copy the compiled `pwru` binary to a new image.

brb:
Fixed by #80

PR #80, opened 2022-08-11
add multi-stage builds to minimize docker image size.

voldemorte:
#78 

The resultant image size is that of about 80MB.

Signed-off-by: Nikhil Chaudhari nikhil.c@mailbox.org

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,30 @@
-FROM ubuntu:20.04
-RUN apt update -y -q
-RUN DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y -q curl build-essential ca-certificates
-RUN curl -s https://storage.googleapis.com/golang/go1.18.3.linux-amd64.tar.gz| tar -v -C /usr/local -xz
+FROM ubuntu:22.04 AS build
+
 ENV PATH $PATH:/usr/local/go/bin
-RUN apt install -y wget gnupg2
-RUN printf "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-12 main" | tee /etc/apt/sources.list.d/llvm-toolchain-xenial-12.list
-RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
-RUN apt -y update
-RUN apt install -y llvm-12 clang-12
-RUN ln -s /usr/bin/clang-12 /usr/bin/clang
-RUN ln -s /usr/lib/llvm-12/bin/llvm-strip /usr/local/bin/llvm-strip
+
+RUN apt update -y -q && \
+    DEBIAN_FRONTEND=noninteractive apt install --no-install-recommends -y -q \
+        curl \
+        build-essential \
+        ca-certificates \
+        wget \
+        gnupg2 \
+        git && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -14,5 +14,8 @@ WORKDIR /pwru
 COPY . .
 RUN make
 RUN chmod a+x /pwru
-ENTRYPOINT ["./pwru"]
-CMD ["./pwru", "-output-tuple"]
+
+FROM ubuntu:20.04 AS base
+COPY --from=build /pwru/pwru /usr/local/bin/
+ENTRYPOINT ["/usr/local/bin/pwru"]
+CMD ["/usr/local/bin/pwru", "-output-tuple"]
```
brb, line 21: Let's drop `CMD`. AFAIK, this then would just run `pwru`.

brb (changes requested):
Thanks! Small nit.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:20.04 AS build
```
brb, line 1: While here, we can s/20.04/22.04/.

Dockerfile
voldemorte: Yes dropping CMD would just run the ENTRYPOINT. Have removed CMD.

Dockerfile
```diff
@@ -14,5 +14,7 @@ WORKDIR /pwru
 COPY . .
 RUN make
 RUN chmod a+x /pwru
-ENTRYPOINT ["./pwru"]
-CMD ["./pwru", "-output-tuple"]
+
+FROM ubuntu:22.04 AS base
```
brb, line 18:
```suggestion
FROM scratch
```
This reduces the final image size to 4MB.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:22.04 AS build
```
brb, line 1: Could you add `git` to `RUN apt install -y wget gnupg2`? This will allow us to add the proper version to the final binary.

brb (changes requested):
Two more comments. Also, could you squash all your commits into single one, and add `Signed-off-by` to the final commit msg?

Dockerfile
```diff
@@ -1,4 +1,4 @@
...
 RUN apt update -y -q
```
tklauser, line 2: While at it, could you merge these `RUN` instructions (i.e. `RUN apt update -y -q && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y -q curl build-essential ca-certificates ...`) to reduce the number of layers (see https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers) as well?

Dockerfile
voldemorte: Yes this definitely reduced the size of the docker image. This was something new I learnt today, Thanks. :)

Dockerfile
voldemorte: I've added `git` to the `apt install`.

Dockerfile
voldemorte: I've managed to club the different `RUN` steps to reduce the number of layers.

brb (approved):
Thanks!

PR #86, opened 2022-09-01
Try to fix endbr64 kprobe mess

### LINKED ISSUE #72: Func name resolution is broken on 5.18
On 5.18, both `PT_REGS_IP(ctx)` and `bpf_get_func_ip(ctx)` return addrs which are +4 compared to 5.17. This breaks IP => func name resolution via `/proc/kallsyms`.

@YutaroHayakawa found this commit https://github.com/torvalds/linux/commit/7f0059b58f0257d895fafd2f2e3afe3bbdf21e64 which might explain the +4.

To apply the same fix we might need to teach https://github.com/cilium/ebpf how to load the kernel config.

brb:
On 5.18+ when CONFIG_X86_KERNEL_IBT is set *some* functions gets endbr64 instruction (size 4 bytes) which screws up the kprobe addressing \[1\].

Assume that if no kallsym is found, then it's due to the endbr64.

Tested locally:

    $ iptables -t filter -I OUTPUT 1 -m tcp --proto tcp \
                --dst 1.1.1.1/32 -j DROP
    $ pwru --filter-dst-ip=1.1.1.1 --output-tuple > foo&
    $ curl 1.1.1.1
    $ cat foo
	PROCESS                     FUNC
	 [curl]             ip_local_out
	 [curl]           __ip_local_out
	 [curl]             nf_hook_slow
	 [curl]         kfree_skb_reason
	 [curl]   skb_release_head_state
	 [curl]                tcp_wfree
	 [curl]         skb_release_data
	 [curl]             kfree_skbmem

\[1\]: https://lore.kernel.org/bpf/20220811091526.172610-5-jolsa@kernel.org/

Fix #72

### DIFF
--- a/internal/pwru/output.go
+++ b/internal/pwru/output.go
@@ -63,6 +63,11 @@ func (o *output) Print(event *Event) {
 	var funcName string
 	if ksym, ok := o.addr2name.Addr2NameMap[addr]; ok {
 		funcName = ksym.name
+	} else if ksym, ok := o.addr2name.Addr2NameMap[addr-4]; runtime.GOARCH == "amd64" && ok {
+		// Assume that function has ENDBR in its prelude (enabled by CONFIG_X86_KERNEL_IBT).
+		// See https://lore.kernel.org/bpf/20220811091526.172610-5-jolsa@kernel.org/
+		// for more ctx.
+		funcName = ksym.name
 	} else {
 		funcName = fmt.Sprintf("0x%x", addr)
 	}


internal/pwru/output.go
```diff
@@ -63,6 +63,11 @@ func (o *output) Print(event *Event) {
 	var funcName string
 	if ksym, ok := o.addr2name.Addr2NameMap[addr]; ok {
 		funcName = ksym.name
+		// Assume that function has ENDBR in its prelude (enabled by CONFIG_X86_KERNEL_IBT).
+		// See https://lore.kernel.org/bpf/20220811091526.172610-5-jolsa@kernel.org/
+		// for more ctx.
+	} else if ksym, ok := o.addr2name.Addr2NameMap[addr-4]; runtime.GOARCH == "amd64" && ok {
```
tklauser, line 69: Very minor nit: I assume the comment applies to the `else if` branch. Might make sense to move it within the branch, otherwise it might be considered to be part of the `if` branch.

```suggestion
	} else if ksym, ok := o.addr2name.Addr2NameMap[addr-4]; runtime.GOARCH == "amd64" && ok {
		// Assume that function has ENDBR in its prelude (enabled by CONFIG_X86_KERNEL_IBT).
		// See https://lore.kernel.org/bpf/20220811091526.172610-5-jolsa@kernel.org/
		// for more ctx.
```

tklauser (approved):

tklauser (approved):

PR #96, opened 2022-09-22
Update Dockerfile to use llvm-14

jauderho:
This updates the Dockerfile to use llvm-13 instead.

Signed-off-by: Jauder Ho <jauderho@users.noreply.github.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,16 +9,10 @@ RUN apt update -y -q && \
         ca-certificates \
         wget \
         gnupg2 \
-        git && \
-    curl -s https://storage.googleapis.com/golang/go1.18.3.linux-amd64.tar.gz| tar -v -C /usr/local -xz && \
-    printf "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-12 main" | tee /etc/apt/sources.list.d/llvm-toolchain-xenial-12.list && \
-    wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
-    apt -y update && \
-    apt install --no-install-recommends -y -q \
-        llvm-12 \
-        clang-12 && \
-    ln -s /usr/bin/clang-12 /usr/bin/clang && \
-    ln -s /usr/lib/llvm-12/bin/llvm-strip /usr/local/bin/llvm-strip
+        git \
+        llvm \
+        clang && \
+    curl -s https://storage.googleapis.com/golang/go1.19.1.linux-amd64.tar.gz | tar -v -C /usr/local -xz
 
 WORKDIR /pwru
 COPY . .


Dockerfile
```diff
@@ -9,16 +9,11 @@ RUN apt update -y -q && \
         ca-certificates \
         wget \
         gnupg2 \
-        git && \
-    curl -s https://storage.googleapis.com/golang/go1.18.3.linux-amd64.tar.gz| tar -v -C /usr/local -xz && \
```
tklauser, line 13: We need to keep this line, otherwise we won't have the `go` binary available.

Dockerfile
```diff
@@ -9,16 +9,11 @@ RUN apt update -y -q && \
...
-    apt install --no-install-recommends -y -q \
-        llvm-12 \
-        clang-12 && \
-    ln -s /usr/bin/clang-12 /usr/bin/clang && \
-    ln -s /usr/lib/llvm-12/bin/llvm-strip /usr/local/bin/llvm-strip
+        git \
+        llvm-13 \
+        clang-13 && \
+    ln -s /usr/bin/clang-13 /usr/bin/clang && \
+    ln -s /usr/lib/llvm-13/bin/llvm-strip /usr/local/bin/llvm-strip
```
tklauser, line 16: Installing the `llvm` and `clang` packages (currently pointing to LLVM 14) instead which will take care of installing the respective symlinks in /usr/bin as well?

```suggestion
        llvm \
        clang
```

tklauser (changes requested):

Dockerfile
jauderho: Should this just be replaced with `golang`?

tklauser: Let's keep using the binary distribution. We could also update to 1.19.1 while at it.

jauderho (commented):
Done

Dockerfile
```diff
@@ -9,16 +9,10 @@ RUN apt update -y -q && \
...
-    printf "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-12 main" | tee /etc/apt/sources.list.d/llvm-toolchain-xenial-12.list && \
-    wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
-    apt -y update && \
-    apt install --no-install-recommends -y -q \
-        llvm-12 \
-        clang-12 && \
-    ln -s /usr/bin/clang-12 /usr/bin/clang && \
-    ln -s /usr/lib/llvm-12/bin/llvm-strip /usr/local/bin/llvm-strip
+        git \
+        curl -s https://storage.googleapis.com/golang/go1.19.1.linux-amd64.tar.gz | tar -v -C /usr/local -xz && \
```
tklauser, line 13: The `curl` command needs to be a separate command rather than part of the `apt install` command package list.

tklauser (changes requested):

Dockerfile
jauderho: My bad, not enough coffee. 

I just pushed a new change which puts the && in the right place too.

tklauser (approved):

PR #98, opened 2022-09-27
Filter out non-available_filter_functions

brb:
It is possible to attach the kprobes only to those functions which are listed in /sys/kernel/debug/tracing/available_filter_functions \[1\].

Previously, we tried to attach the kprobes to all functions. Attaching to a non-listed function failed with ENOENT which was not a problem, as we simply ignored such an error.

However, this becomes a problem with an introduction of the multi-link kprobes mechanism, as any invalid syms makes attachment to all other valid syms to fail.

\[1\]: https://www.kernel.org/doc/Documentation/trace/ftrace.rst

cc @vincentmli 

---

NB: it's non-trivial to fix the K8s example. I'm planning to do `mount -t debugfs none /sys/kernel/debug/` from `pwru`, so that no extra mounts are needed for K8s / Docker.

### DIFF
--- a/internal/pwru/utils.go
+++ b/internal/pwru/utils.go
@@ -5,7 +5,9 @@
 package pwru
 
 import (
+	"bufio"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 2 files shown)


vincentmli:
@brb sort of half related question, we have 1) function not in available filter function 2) duplicate symbols provided to kernel.
I tested 2) will also cause kernel return ESRCH, 2) is not possible with pwru, right?

brb:
Yeah, duplicate symbols are not possible.

vincentmli:
ok, cool, I can rebase my PR once this is merged, it will be much easier :)

internal/pwru/utils.go
```diff
@@ -15,15 +17,45 @@ import (
...
 
+// getAvailableFilterFunctions return list of functions to which it is possible
+// to attach kprobes.
+func getAvailableFilterFunctions() (map[string]struct{}, error) {
+	availableFuncs := make(map[string]struct{})
+	path := "/sys/kernel/debug/tracing/available_filter_functions"
+
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open %s: %v", path, err)
```
tklauser, line 28: Nit: I think we can return the plain `err` here, `os.Open` will return an `os.PathError` which already contains the path.

```suggestion
		return nil, err
```

internal/pwru/utils.go
```diff
@@ -15,15 +17,45 @@ import (
...
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	scanner.Split(bufio.ScanLines)
+	for scanner.Scan() {
+		availableFuncs[scanner.Text()] = struct{}{}
+	}
+
```
tklauser, line 37: Let's check `scanner.Err()` here:

```suggestion
if err := scanner.Err(); err != nil {
	return nil, err
}
```

tklauser (approved):
Thanks

PR #99, opened 2022-09-28
Add kprobe multi-link support

### LINKED ISSUE #87: Use multi-link kprobe attachment
Once https://github.com/cilium/ebpf/pull/716 has been merged, we can start using the multi-link kprobes. This should significantly reduce the startup time.

brb:
This makes the kprobes loading blazingly fast!

cc @vincentmli

Fix #87

### DIFF
--- a/internal/pwru/types.go
+++ b/internal/pwru/types.go
@@ -5,6 +5,7 @@
 package pwru
 
 import (
+	"fmt"
 	"os"
 
 	flag "github.com/spf13/pflag"
@@ -12,6 +13,9 @@ import (
 
 const (
 	MaxStackDepth = 50
+
+	BackendKprobe      = "kprobe"
+	BackendKprobeMulti = "kprobe-multi"
 )
 
 type Flags struct {
@@ -41,6 +45,8 @@ type Flags struct {
 	AllKMods     bool
 
 	ReadyFile string
+
+	Backend string
 }
 
 func (f *Flags) SetFlags() {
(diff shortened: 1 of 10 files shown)


main.go
```diff
@@ -82,7 +82,20 @@ func main() {
 		}
 	}
 
-	funcs, err := pwru.GetFuncs(flags.FilterFunc, btfSpec, flags.KMods)
+	var useKprobeMulti bool
+	if flags.Backend != "" && (flags.Backend != "kprobe" && flags.Backend != "kprobe-multi") {
```
tklauser, line 86: Nit: maybe move the `"kprobe"` and `"kprobe-multi"` strings into consts, so they could be reused below and in the usage message for `--backend`?

internal/pwru/types.go
```diff
@@ -68,6 +70,8 @@ func (f *Flags) SetFlags() {
 
 	flag.StringVar(&f.ReadyFile, "ready-file", "", "create file after all BPF progs are attached")
 	flag.Lookup("ready-file").Hidden = true
+
+	flag.StringVar(&f.Backend, "backend", "", "Tracing backend('kprobe', 'kprobe-multi')")
```
tklauser, line 74: Maybe worth mentioning the the usage that this is auto-detected in case backend is empty and kmods are empty. At least that's how I read the following:

https://github.com/cilium/pwru/pull/99/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R92-R93

PR #104, opened 2022-10-11
Update docker/Kubernetes deployment instructions in README

anubhabMajumdar:
Using the current `Running pwru with Docker` command throws the following error in Ubuntu 22.04/22.10

```
$> docker run --privileged --rm -t --pid=host -v /sys/kernel/debug/tracing:/sys/kernel/debug/tracing cilium/pwru --filter-dst-ip=1.1.1.1
    docker: Error response from daemon: error while creating mount source path '/sys/kernel/debug/tracing': mkdir /sys/kernel/debug/tracing: no such file or directory.
```

This is similar to the issue reported here: https://forums.docker.com/t/mount-subpath-of-debugfs/123525

One way to resolve this would be to mount the entire `/sys/kernel/debug` to the `pwru` container.

Similar error is surfaced when applying the Kubernetes spec mentioned in the README. I could get `pwru` running on Kubernetes node only after creating and attaching volume.

Signed-off-by: Anubhab Majumdar <anmajumdar@microsoft.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,15 @@ An example how to run `pwru` with Docker:
 docker run --privileged --rm -t --pid=host -v /sys/kernel/debug/tracing:/sys/kernel/debug/tracing cilium/pwru --filter-dst-ip=1.1.1.1
 ```
 
+Note: In case the path `/sys/kernel/debug/tracing` doesn't exist in the container, you can see error similar to 
+```
+docker: Error response from daemon: error while creating mount source path '/sys/kernel/debug/tracing': mkdir /sys/kernel/debug/tracing: no such file or directory
+```
+One way to avoid this is to mount the entire `/sys/kernel/debug` path from host to the container.
+```
+docker run --privileged --rm -t --pid=host -v /sys/kernel/debug:/sys/kernel/debug cilium/pwru --filter-dst-ip=1.1.1.1
+```
+
 ### Running on Kubernetes
 
 The following example shows how to run `pwru` on a given node:
@@ -88,6 +97,8 @@ kubectl run pwru \
     --overrides='{"apiVersion":"v1","spec":{"nodeSelector":{"kubernetes.io/hostname":"'$NODE'"}, "hostNetwork": true, "hostPID": true}}' \
     -- --filter-dst-ip=1.1.1.1 --output-tuple
 ```
+
+Note: You may need to create a volume for `/sys/kernel/debug/tracing` and mount it for the`pwru` pod. Similar to docker, if `/sys/kernel/debug/tracing` path doesn't exist in the container, mount entire `/sys/kernel/debug` to the container.
 
 ### Running on Vagrant
 


README.md
```diff
@@ -73,20 +73,48 @@ Docker images for `pwru` are published at https://hub.docker.com/r/cilium/pwru.
 An example how to run `pwru` with Docker:
 
 ```
-docker run --privileged --rm -t --pid=host -v /sys/kernel/debug/tracing:/sys/kernel/debug/tracing cilium/pwru --filter-dst-ip=1.1.1.1
```
aditighag, line 76: `/sys/kernel/debug/tracing` is a standard path for tracing on Linux platform. Have you looked into enabling tracing on your machine?

anubhabMajumdar: Tracing is enabled on my machine, and I checked that `/sys/kernel/debug/tracing` exists on the host machine. I don't think this path exists on the ubuntu:22.04 container. I can use the `pwru` binary on my machine, as well as use the docker image when I mount `/sys/kernel/debug` instead of `/sys/kernel/debug/tracing`.

anubhabMajumdar: I can edit the review to add this as a note. Will that be better?

aditighag: >I checked that /sys/kernel/debug/tracing exists on the host machine. I don't think this path exists on the ubuntu:22.04 container.

So I'm not sure where the issue is. The command in the README mounts the tracing directory from the host inside the pwru container. So what do you mean by "I don't think this path exists on the ubuntu:22.04 container"?

anubhabMajumdar: `/sys/kernel/debug` is RO, and docker cannot create `/sys/kernel/debug/tracing` directory to complete the mount operation for `pwru` container. If the path `/sys/kernel/debug/tracing` existed, then the given command would succeed.  Also, cannot mount this to any other location in the container, because the code checks this path.
```
$ docker run --privileged --rm -t --pid=host -v /sys/kernel/debug/tracing:/sys/kernel/debug cilium/pwru --filter-dst-ip=1.1.1.1
2022/10/11 22:02:43 Failed to retrieve available ftrace functions (is /sys/kernel/debug/tracing mounted?): failed to open: open /sys/kernel/debug/tracing/available_filter_functions: no such file or directory
...
```

Hence, I think it may be worth noting in the README that mounting `/sys/kernel/debug` entirely could act as a workaround.

aditighag: The note looks reasonable to me. Thanks!

README.md
```diff
@@ -89,6 +98,47 @@ kubectl run pwru \
     -- --filter-dst-ip=1.1.1.1 --output-tuple
 ```
 
+Note: You may need to create a volume for `/sys/kernel/debug/tracing` and mount it for the`pwru` pod.
```
aditighag, line 101: I'm not sure if we need to include the full pod spec in README. Is it possible to extend the above command @L92 with volume mounts?

README.md
```diff
@@ -89,6 +98,47 @@ kubectl run pwru \
...
+      args:
+        ["--filter-dst-ip=1.1.1.1", "--output-tuple"]
+  volumes:
+    - name: tracing
+      hostPath:
+        path: /sys/kernel/debug/tracing
+        type: Directory
+```
+
+Similar to docker, if `/sys/kernel/debug/tracing` path doesn't exist inh the container, mount entire `/sys/kernel/debug` to the container.
```
brb, line 134: Nit: inh -> in.

brb (changes requested):
Thanks! Small nit.

brb:
@anubhabMajumdar :wave: do you plan to address the feedback?

anubhabMajumdar:
> @anubhabMajumdar 👋 do you plan to address the feedback?

Sorry for the delay; will get to this today. Thanks!

README.md
anubhabMajumdar: Done.

brb:
@anubhabMajumdar Thanks! Could you squash all your commits into one? Then we are ready to merge.

anubhabMajumdar:
> @anubhabMajumdar Thanks! Could you squash all your commits into one? Then we are ready to merge.

Done

brb (approved):

Issue #100, opened 2022-09-29
Add flag to attach to all available kernel modules


lbernail:
We can do this today with
```--kmods $(ls /sys/kernel/btf/ | grep -v vmlinux | tr '\n' , | sed 's/.$//')```

A `--all-kmods` flag would be nice

**Comment by varunkumare99** [comment]
Hi
I'm new to ebpf and pwru.
I would like to work on this.
Any idea on how to begin?
Thanks

**Comment by brb** [comment]
@varunkumare99 :wave: This would require adding a new flag, and then passing all kernel module names from `/sys/kernel/btf/` to the `GetFuncs()` function. No eBPF knowledge is required. Please let me know if you are willing to work on this, so that I can assign the issue to you.

**Comment by varunkumare99** [comment]
Yes, I'm willing to work on this.

**Comment by varunkumare99** [comment]
I'm unable to run pwru with docker.

> sudo docker run --privileged --rm -t --pid=host -v /sys/kernel/debug/tracing:/sys/kernel/debug/tracing cilium/pwru --filter-dst-ip=1.1.1.1

`docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/sys/kernel/debug/tracing" to rootfs at "/sys/kernel/debug/tracing": mkdir /var/lib/docker/overlay2/c1572db5282c23f07099badddd4e2eabdc3afddc21f8ebd42be43faf357cd635/merged/sys/kernel/debug/tracing: no such file or directory: unknown.`

**Comment by brb** [comment]
For posterity, the fix was to run with:
```
sudo docker run --privileged --rm -t --pid=host -v /sys/kernel/debug/:/sys/kernel/debug/ cilium/pwru --filter-dst-ip=1.1.1.1
```

varunkumare99:
thanks I'm able to run it now.

varunkumare99:
I have raised a PR : #109

PR #109, opened 2022-11-02
Add flag to attach to all available kernel modules

varunkumare99:
added flag `--all-kmods` to attach to all available kernel modules.
fix: #100

### DIFF
--- a/main.go
+++ b/main.go
@@ -68,6 +68,20 @@ func main() {
 		log.Fatalf("Failed to load BTF spec: %s", err)
 	}
 
+	if flags.AllKMods {
+		files, err := os.ReadDir("/sys/kernel/btf")
+		if err != nil {
+			log.Fatalf("Failed to read directory: %s", err)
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -68,6 +68,20 @@ func main() {
...
 
+	if flags.AllKMods {
+		files, err := os.ReadDir("/sys/kernel/btf")
+		if err != nil {
+			log.Fatalf("Failed to read Directory: %s", err)
+		}
+
+		flags.KMods = nil
+		for _, file := range files {
+			if !file.IsDir() {
```
brb, line 79: You want to exclude `vmlinux` here. Otherwise, it will be parsed twice.

brb (changes requested):
Thanks! LGTM, just two minor comments.

brb (approved):
Thanks!

varunkumare99:
Thank you!

PR #127, opened 2022-12-19
Replace kprobe-shim reflection with generated interfaces

gandro:
This PR replaces the reflection-based approach to obtain the map and program references with an interface-based one. 

The methods defined by the interfaces here are generated via the `getgetter` tool introduced in PR as well. It is executed via `go generate` and generates getters for each exported field in the provided structs. The `getgetter` utility relies on the `goimports` library to format the generated code and fix up any missing imports.


brb (approved):
Thanks!

PR #141, opened 2023-01-11
Add port filter option to allow filtering on either src or dst port

msherif1234:
adding --filter-port to allow filtering or either source or destination port
which helps on debugging bidirectional TCP/UDP flows.
Signed-off-by: Mohamed Mahmoud <mmahmoud@redhat.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -85,13 +85,14 @@ jobs:
 
             # Block traffic to both endpoints, so that we can see kfree_skb*
             # in traces (test assertion)
-            iptables -I OUTPUT 1 -m tcp --proto tcp --dst 1.0.0.1/32 -j DROP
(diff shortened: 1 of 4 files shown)


brb (changes requested):
Thanks! 

1. Shall we make `--filter-{dst,src}-port` mutually exclusive with `--filter-port`?
2. Could you extend the tests to include `--port` ([this](https://github.com/cilium/pwru/blob/main/.github/workflows/test.yml#L88) and [this](https://github.com/cilium/pwru/blob/main/.github/workflows/test.yml#L89))?

msherif1234:
@brb I couldn't find out how to run test locally I tried `make release-local` but I don't think that run it ?

msherif1234:
@brb is there a way for me to trigger the ci test or u have to trigger it for me ?

brb (approved):

PR #148, opened 2023-02-01
Print drop reason from kfree_skb_reason

### LINKED ISSUE #54: Print packet drop reason collected from `kfree_skb_reason`
Details are in the lwn article - https://lwn.net/Articles/885729/.

>This information is not actually useful to the kernel, but it has been added to the existing kfree_skb tracepoint, making it available to any program that connects to that tracepoint.

It looks like the drop reasons are available with the existing trace point.

brb:
This commit extracts a drop reason from kfree_skb_reason(), and prints
it when kfree_skb_reason() is called.

The helper was introduced in the v5.17. The second function param of it
contains a drop reason which is the "skb_drop_reason" enum. To extract
enum, we parse the kernel's BTF, as the enum is not considered as UAPI,
and there were already some position changes among its values.

An example output:

    0xffff936378692ce8 6 [curl]             nf_hook_slow 172.20.10.2:33918->1.1.1.1:4240(tcp)
    0xffff936378692ce8 6 [curl] kfree_skb_reason(SKB_DROP_REASON_NETFILTER_DROP) 172.20.10.2:33918->1.1.1.1:4240(tcp)

An alternative approach to passing the second param would have been to
set a BPF cookie when attaching the kprobe to the kfree_skb_reason().
However, that would require creating a new bpf/kprobe_pwru*.c, as
the helper bpf_get_attach_cookie() requires the bpf_link support, and
thus a newer kernel than the v5.3 (min required by pwru).

Fix #54

### DIFF
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -62,6 +62,7 @@ struct event_t {
 	struct skb_meta meta;
 	struct tuple tuple;
 	s64 print_stack_id;
+	u64 param_second;
 	u32 cpu_id;
 } __attribute__((packed));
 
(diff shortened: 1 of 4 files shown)


brb:
> Did you use this to debug any issues yet?

Yep, this one https://github.com/cilium/cilium/issues/23283 (got `SKB_DROP_REASON_NO_SOCKET` which guided me to a correct direction).

I still need to see how stable is that enum across different kernel versions :crossed_fingers:

tklauser (approved):
Nice!

PR #151, opened 2023-02-06
output: add absolute time to timestamp flag

### LINKED ISSUE #149: Add datetime to the output
Would be awesome to add, as optional, the current datetime for each event line that pwru emits, instead of
```
0xffff8d36e2d97000     11          [agent]         skb_release_data             1616 netns=0 mark=0x0 ifindex=0 proto=8 mtu=0 len=120                                                                                                                          
0xffff8d36e2d97000     11          [agent]             kfree_skbmem             1634 netns=0 mark=0x0 ifindex=0 proto=8 mtu=0 len=120                                                                                                                          
0xffff8d36e2d96200     11          [agent]          skb_consume_udp         10517284 netns=0 mark=0x0 ifindex=0 proto=8 mtu=0 len=116                                                                                                                          
0xffff8d36e2d96200     11          [agent]  __consume_stateless_skb             1194 netns=0 mark=0x0 ifindex=0 proto=8 mtu=0 len=116                                                                                                                          
0xffff8d36e2d96200     11          [agent]         skb_release_data             1203 netns=0 mark=0x0 ifindex=0 proto=8 mtu=0 len=116 
```
we would have
```
18:03:00 0xffff8d36e2d97000     11          [agent]         skb_release_data             1616 netns=0 mark=0x0 ifindex=0 proto=8 mtu=0 len=120                                                                                                                          
18:03:00 0xffff8d36e2d97000     11          [agent]             kfree_skbmem             1634 netns=0 mark=0x0 ifindex=0 proto=8 mtu=0 len=120                                                                                                                          
18:03:00 0xffff8d36e2d96200     11          [agent]          skb_consume_udp         10517284 netns=0 mark=0x0 ifindex=0 proto=8 mtu=0 len=116                                                                                                                          
18:03:00 0xffff8d36e2d96200     11          [agent]  __consume_stateless_skb             1194 netns=0 mark=0x0 ifindex=0 proto=8 mtu=0 len=116                                                                                                                          
18:03:00 0xffff8d36e2d96200     11          [agent]         skb_release_data             1203 netns=0 mark=0x0 ifindex=0 proto=8 mtu=0 len=116 
```
wdyt @brb?

sterchelen:
This patch adds "absolute" as a new value to the timestamp flag.
When set, it prefixes each line with the current time.

Closes #149

### DIFF
--- a/internal/pwru/output.go
+++ b/internal/pwru/output.go
@@ -11,12 +11,15 @@ import (
 	"os"
 	"runtime"
 	"syscall"
+	"time"
 
 	"github.com/cilium/ebpf"
 	ps "github.com/mitchellh/go-ps"
 
 	"github.com/cilium/pwru/internal/byteorder"
 )
+
+const absoluteTS string = "15:04:05.000"
 
 type output struct {
 	flags         *Flags
@@ -53,6 +56,9 @@ func NewOutput(flags *Flags, printSkbMap *ebpf.Map, printStackMap *ebpf.Map,
 }
 
 func (o *output) PrintHeader() {
+	if o.flags.OutputTS == "absolute" {
+		fmt.Fprintf(o.writer, "%12s ", "TIME")
+	}
 	fmt.Fprintf(o.writer, "%18s %6s %16s %24s", "SKB", "CPU", "PROCESS", "FUNC")
 	if o.flags.OutputTS != "none" {
 		fmt.Fprintf(o.writer, " %16s", "TIMESTAMP")
@@ -61,6 +67,9 @@ func (o *output) PrintHeader() {
 }
 
 func (o *output) Print(event *Event) {
+	if o.flags.OutputTS == "absolute" {
+		fmt.Fprintf(o.writer, "%12s ", time.Now().Format(absoluteTS))
+	}
 	p, err := ps.FindProcess(int(event.PID))
 	execName := "<empty>"
 	if err == nil && p != nil {
--- a/internal/pwru/types.go
+++ b/internal/pwru/types.go
@@ -65,7 +65,7 @@ func (f *Flags) SetFlags() {
 	flag.Uint16Var(&f.FilterSrcPort, "filter-src-port", 0, "filter source port")
 	flag.Uint16Var(&f.FilterDstPort, "filter-dst-port", 0, "filter destination port")
 	flag.Uint16Var(&f.FilterPort, "filter-port", 0, "filter either destination or source port")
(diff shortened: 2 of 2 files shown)


internal/pwru/output.go
```diff
@@ -61,6 +65,9 @@ func (o *output) PrintHeader() {
 }
 
 func (o *output) Print(event *Event) {
+	if o.flags.OutputAbsoluteTimestamp {
+		fmt.Fprintf(o.writer, "%s ", time.Now().Format(time.TimeOnly))
```
brb, line 69: I think it makes sense to add milliseconds too.

internal/pwru/types.go
```diff
@@ -66,6 +67,7 @@ func (f *Flags) SetFlags() {
 	flag.Uint16Var(&f.FilterDstPort, "filter-dst-port", 0, "filter destination port")
 	flag.Uint16Var(&f.FilterPort, "filter-port", 0, "filter either destination or source port")
 	flag.StringVar(&f.OutputTS, "timestamp", "none", "print timestamp per skb (\"current\", \"relative\", \"none\")")
+	flag.BoolVar(&f.OutputAbsoluteTimestamp, "output-absolute-timestamp", false, "prefix each line with the current time")
```
brb, line 70: Instead of creating a new param, could we extend `OutputTS` to accept `"absolute"`?

brb (changes requested):
Thanks for the PR! Some minor comments.

internal/pwru/types.go
sterchelen: How you want to combine the sk_buff timestamp with the absolute one?

brb: I'd suggest either to enable sk_buff timestamp (`current`, `relative`) or enable your user space timestamps (`absolute`) or complete disable (`none`).

sterchelen: Sad to have only the choice between either sk_buff's timestamp or absolute one. I believe that we should keep the possibility to have both. 
What do you think? It's not a strong opinion but a feeling that it deserves a special flag. What are your pov?

brb: I'd be in favor to have as few as possible configuration options. If folks want to use both, then we can always introduce both. But for now I'd suggest to keep it in the same option.

sterchelen: Know that I have the reason, I agree with you to keep as few as possible the number of config 👍🏼. 
Thx for sharing your opinion 🙇🏼.

brb (approved):
Thanks!

PR #157, opened 2023-02-17
Fix icmp6 filter

brb:
Previously, when no IPv6 addr was specified, the ICMPv6 filtering was bypassed.

### DIFF
--- a/internal/pwru/config.go
+++ b/internal/pwru/config.go
@@ -79,18 +79,25 @@ func GetConfig(flags *Flags) FilterCfg {
 		cfg.FilterProto = syscall.IPPROTO_ICMP
 	case "icmp6":
 		cfg.FilterProto = syscall.IPPROTO_ICMPV6
+		cfg.FilterIPv6 = 1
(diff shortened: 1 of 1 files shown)


msherif1234 (changes requested):
At the dst filter check if it’s ipv4 and IPv6 Config is 1 do we need to set version match to false ?

internal/pwru/config.go
```diff
@@ -79,6 +79,7 @@ func GetConfig(flags *Flags) FilterCfg {
 		cfg.FilterProto = syscall.IPPROTO_ICMP
 	case "icmp6":
 		cfg.FilterProto = syscall.IPPROTO_ICMPV6
+		cfg.FilterIPv6 = 1
```
tklauser, line 82: Should we check cfg.FilterIPv6 in the `if flags.FilterDstIP != ""` branch below and bail out in case it is set and the dst IP is an IPv4?

tklauser (approved):

PR #187, opened 2023-06-02
add tcp flag filtering

yuval-k:
Add a new option for filtering using TCP Flags. semantics is similar to iptables. You provide a set of flags and a mask to test the flags with.

I havn't seen existing tests for filtering logic, so would appreciate some insight on how to best tests this. I've ran it manually and it seems to work.

### DIFF
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -77,6 +77,9 @@ struct config {
 	union addr saddr;
 	union addr daddr;
 	u8 l4_proto;
+	u8 tcp_flags;
+	u8 tcp_flags_mask;
+
 	u16 sport;
 	u16 dport;
 	u16 port;
@@ -152,7 +155,7 @@ config_tuple_empty() {
 	if (!addr_empty(cfg->saddr) || !addr_empty(cfg->daddr)) {
 		return false;
 	}
-	if (cfg->l4_proto || cfg->sport || cfg->dport || cfg->port) {
+	if (cfg->l4_proto || cfg->sport || cfg->dport || cfg->port || cfg->tcp_flags_mask) {
 		return false;
 	}
 	return true;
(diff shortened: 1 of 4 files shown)


internal/pwru/config.go
```diff
@@ -41,6 +44,30 @@ type FilterCfg struct {
...
+		return 0x10
+	case "urg":
+		return 0x20
+	case "ece":
+		return 0x40
+	case "cwr":
+		return 0x80
+	default:
+		log.Fatalf("Invalid tcp flag format")
+		panic("unreachable")
```
brb, line 67: Nit: this is redundant, as `log.Fatalf()` does `panic()`.

brb (commented):
Thanks for the contribution!

Regarding testing, I think you could extend the `.github/worklows/test.yaml` and set `--filter-tcp-flags=syn`

internal/pwru/config.go
yuval-k: yes, it is redundant, but the go compiler doesn't know that. if there panic is not there it will fail to build with "missing return" error.

bpf/kprobe_pwru.c
```diff
@@ -236,9 +241,11 @@ filter_l3_and_l4(struct sk_buff *skb) {
 		if (cfg->port && (dport != cfg->port && sport != cfg->port)) {
 			return false;
 		}
+		if (cfg->tcp_flags_mask && ((flags & cfg->tcp_flags_mask) != cfg->tcp_flags)) {
```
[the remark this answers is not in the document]
yuval-k: I do not have example where mask is needed - I added the mask in this PR to make more generic (and hopefully more useful for others), and took inspiration from the way iptables tcp flag matching works today  (`--tcp-flags` in iptables). 

I'm happy to change this, i think we can do one of the following:
- remove mask all together, or
- default the mask to 0xff if not specified

WDYT?

yuval-k:
@brb thank you for the review; I'm a bit under the weather this week, so will be able to make code changes/tests next week

yuval-k:
@brb I updated the GH workflow to include the new flag. Can you please approve the workflow so that CI runs?

brb:
Hey, thanks for the update. I'm a bit short on time this week, will try to re-review by EOW.

bpf/kprobe_pwru.c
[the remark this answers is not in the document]
brb: Considering that we will have a full tcpdump-like filter syntax support soon (https://github.com/cilium/pwru/pull/198), I'd suggest to drop the whole mask.

internal/pwru/config.go
brb: Let's return 0 instead of panic.

PR #195, opened 2023-06-21
Replace BPF_MAP_TYPE_PERF_EVENT_ARRAY by BPF_MAP_TYPE_QUEUE

jschwinger233:
BPF_MAP_TYPE_QUEUE has advantages over BPF_MAP_TYPE_PERF_EVENT_ARRAY:
1. Event ordering: thanks to bpf queue's shared buffer among CPUs and FIFO data structure, there wouldn't be an out of order issue for events.
2. Event loss: perf buffer is so small that it's common to drop events, which causes trouble in tracing some specific critical datapath, such as checking skb->mark after tcf_classify. Bpf queue won't lose a simple event until it's filled up to capacity.

Another option is BPF_MAP_TYPE_RINGBUF, but it doesn't solve the issue of dropping events, and it has even worse performance when called with flag `BPF_RB_FORCE_WAKEUP`.

Our concerns are performance and memory pressure:

For performance, I created a benchmark repo for
BPF_MAP_TYPE_PERF_EVENT_ARRAY, BPF_MAP_TYPE_QUEUE, and BPF_MAP_TYPE_RINGBUF at
https://github.com/jschwinger233/bpf_k2u_benchmark. The results showed bpf queue is approximate 1.5X ~ 2X slower than perf out, while ringbuf output is rather unstable and sometimes is even worst-performing.

For memory pressure, bpf queue requires memory allocation for sizeof(event) * capacity. For pwru, it's 124 bytes * 10,000 = 1.24 Mb.

I understand it's frustrating to face the performace setback, especically when pwru is used in the environment with busy traffic, which could be risky to drag down the system. After all, it's a tradeoff to have either better observability or better performance.


brb (approved):
Thanks!

Issue #142, opened 2023-01-12
IPv6 L4 Protocol


ur4t:
```
/*
 * TODO: ipv6 l4 protocol
 * According to https://www.rfc-editor.org/rfc/rfc2460, in ipv6 header, the
 * transport layer protocol is represented by the Next Header field. However
 * ipv6 supports extension headers and recommends to place the transport layer
 * protocol at last. So if we want to parse out the transport layer protocol,
 * we have to identify all the extension headers, which is quite troublesome.
 * Currently it is assumed that there are no ipv6 extension headers.
 */
```

brb:
It will be addressed by #198.

Issue #85, opened 2022-08-26
support --filter-ip and --filter-port


borkmann:
Right now pwru only supports:
```
      --filter-dst-ip string        filter destination IP addr
      --filter-dst-port uint16      filter destination port
[...]
      --filter-src-ip string        filter source IP addr
      --filter-src-port uint16      filter source port
```
But when tracing packets through the stack, they can get DNAT'ed or SNAT'ed or both.

Maybe libpcap-like expression could even be used here, and converted to eBPF insns (e.g. inline asm blob), at least basic primitives to define an expression with ips & ports combined with logical and/or.

Asphaltt:
It's interesting to compile libpcap-like expression to JIT'ed eBPF insns, and make them work with `pwru` existing eBPF insns.

How about `bpf_tail_call()`? Get the `skb` first, then `bpf_tail_call()` the JIT'ed filtering the `skb` insns, and then `bpf_tail_call()` the `pwru` filtering and outputing eBPF insns.

brb:
> It's interesting to compile libpcap-like expression to JIT'ed eBPF insns, and make them work with pwru existing eBPF insns.

Yeah, in my spare time I am working on a small tcpdump-like syntax compiler to eBPF. Instead of `bpf_tail_call()` the plan is to use bpf2bpf calls.

Asphaltt:
@brb  How is it going?

>  Instead of bpf_tail_call() the plan is to use bpf2bpf calls.

After some researching, `bpf2bpf` + `freplace` is perfect for this case.

brb:
@Asphaltt Hey, this is still WIP. ACK on `bpf2bpf` + `freplace`.

PR #198, opened 2023-06-22
Use libpcap-filter for packet filtering

jschwinger233:
> This is a POC PR to demonstrate how to introduce libpcap for pwru, will clean this shit up after some detail discussions.
> 
> 1. Statically link for libpcap. It's feasible, as long as we compile the libpcap.a with some specific compilation flags, which can be integrated into release flow. This PR doesn't implement this for now, but we won't be bothered by the linking trouble.
> 2. Generating filter ebpf instructions. 1) libpcap API compiles pcap-filter expression into cbpf; 2) cloudflare/cbpfc compiles cbpf into ebpf; 3) we, pwru, translate direct skb fields accessing into `bpf_probe_read_kernel`. The last step is very subtle and this PR still leaves some issue to make it work perfectly, see the `TODO@gray` in the PR.
> 3. Inject generated ebpf instructions into existing instructions. To achieve this task, I add a `bpf_printk` in the bpf c code, which is exactly the position I will inject the filter instructions. Adding bpf_printk is good because: 1) it tells us the location of skb->data (L3 header pointer) and skb->data_end (...), which is necessary for cloudflare/cbpfc to do the conversion; 2) it leaves R1, R2, R3, R4 available for cloudflare/cbpfc conversion; 3) it's a clear signal to let us find the injection point: we just need to find the instruction calling bpf_printk; Still there are error-prone parts: we have to adjust offsets for some jump instructions.
> 4. For the legacy filter flags such as `--filter-dst-ip`, we can make it part of the pcap-filter expression. For example, if user run pwru with `--filter-dst-ip 1.2.3.4 --pcap-filter 'tcp and dst port 80'`, we can compose a new pcap filter expression: `tcp and dst port 80 and dest host 1.2.3.4`. So all the existing filtering code in bpf c becomes obsolete. However, this PR hasn't done this so far, but the point is we won't break the CLI interface.
> 
> Introducing tailcall or similar mechanism to decouple the generated instructions from the existing part is valuable, but I decided to implement this POC first to see if there is some hidden obstacle.
> 
> Current implementation is buggy and dirty, full of debugging logs, but I can improve it later if the ideas look good to @brb 
> I know my explanation is still vague, feel free to ask anything.

This PR allows pwru to filter traffic using [pcap-filter(7)](https://www.tcpdump.org/manpages/pcap-filter.7.html) expression. For example, we can run `pwru --filter-pcap 'ip6[40+13] & 0x3f = 0x2'` to capture Syn TCP on IPv6.

The main ideas to achieve the goal are:
1. Link libpcap, use its API to compile pcap-filter expression to cbpf
2. Use cloudflare/cbpfc to convert cbpf to ebpf
3. Adjust generated ebpf instructions to pass the verifier
4. Inject the instructions to a specific position

Although the steps sound superficial and simple, the implementation details require meticulousness. See the comments in the libpcap package.

The original CLI flags won't be broken, as we converted them into part of pcap-filter expression.

Fixes: #85 
Fixes: https://github.com/cilium/pwru/issues/142


brb:
Very nice!

Answers to the points above:

1. Yep, we can embed libpcap.a at a later stage.
2. My main worry about relying on `cloudflare/cbpfc` is that it introduces a runtime dependency on Clang (or did I completely misread their README?)
3. SGTM.
4. Yep, I'd bother about the conversion for now.
5 (the tail call). Yep, we can do it later. Either a tail call or freplace could do the job.

brb:
> My main worry about relying on cloudflare/cbpfc is that it introduces a runtime dependency on Clang (or did I completely misread their README?)

@jibi instructed me that it's only for C compilation. So please ignore my concern.

jschwinger233:
Functionality should be fine now, next I can clean the code up and make it a readable PR.

brb:
> Functionality should be fine now, next I can clean the code up and make it a readable PR.

:+1: Next release is going to be crazy good.

jschwinger233:
@brb already for review but don't merge until I fix the release flow😂

Makefile
```diff
@@ -1,5 +1,5 @@
 GO := go
-GO_BUILD = CGO_ENABLED=0 $(GO) build
+GO_BUILD = CGO_ENABLED=1 $(GO) build
```
brb, line 2: Shall we add `-ldflags "-linkmode 'external' -extldflags '-static'"` to get a statically linked exec?

.github/workflows/test.yml
```diff
@@ -24,6 +24,10 @@ jobs:
         go mod vendor
         go mod verify
         test -z "$(git status --porcelain)" || (echo "please run 'go mod tidy && go mod vendor', and submit your changes"; exit 1)
+        sudo apt-get install -y curl unzip gcc flex bison make
```
brb, line 27: Could we add this step to Makefile under libpcap.a target?

Makefile
jschwinger233: It's in the `libpcap/compile.go`, so `go build` without ldflags can generate statically linked elf:

```go
/*
#cgo LDFLAGS: -L/usr/local/lib -lpcap -static
#include <stdlib.h>
#include <pcap.h>
*/
import "C"
```

Or it's better if we move the flags to `go build`?

brb: TIL that `-static` in `#cgo` forces building a statically linked go exec. It's fine to keep it here.

libpcap/compile.go
```diff
@@ -0,0 +1,202 @@
...
+	*/
+	insts = append([]asm.Instruction{
+		asm.Mov.Imm(asm.R1, 0),
+		asm.Mov.Imm(asm.R2, 0),
+		asm.Mov.Imm(asm.R3, 0),
+	}, insts...)
+
+	// Append instructions to implement "exit immediately if not matched"
+	insts = append(insts,
+		asm.Mov.Imm(asm.R0, 0).WithSymbol("result"), // r0 = TC_ACT_OK
```
brb, line 195: Nit: s/TC_ACT_OK/0/

libpcap/inject.go
```diff
@@ -0,0 +1,168 @@
...
+4. Inject the instructions
+*/
+func InjectFilter(program *ebpf.ProgramSpec, filterExpr string) (err error) {
+	if filterExpr == "" {
+		return
+	}
+
+	injectIdx := 0
+	for idx, inst := range program.Instructions {
+		// In the kprobe_pwru.c, we deliberately put a bpf_printk call to mark the injection position, see the comments over there.
```
brb, line 26: GL to folks who will use bpf_printk for debugging :-) Maybe let's add a warning as a code comment in the beginning of kprobe_pwru.c?

libpcap/inject.go
```diff
@@ -0,0 +1,168 @@
...
+
+/*
+Steps:
+1. Find the injection position, which is the bpf_printk call
+2. Make some necessary preparations for the injection
+3. Compile the filter expression into ebpf instructions
+4. Inject the instructions
+*/
+func InjectFilter(program *ebpf.ProgramSpec, filterExpr string) (err error) {
+	if filterExpr == "" {
```
brb, line 20: If empty filter, then we will end up calling the bpf_printk?

libpcap/inject.go
```diff
@@ -0,0 +1,168 @@
...
+		*/
+		if inst.OpCode.Class().IsJump() {
+			// Zero jump offset implies a function call, leave it alone
+			if inst.Offset == 0 {
+				continue
+			}
+
+			// If there already is a reference and corresponding
+			// symbol, we don't have to create new symbol, just set -1
+			// to the offset so that cilium/ebpf loader can adjust it.
```
brb, line 64: TIL about this loader's feature :star_struck:

internal/pwru/types.go
```diff
@@ -34,6 +34,7 @@ type Flags struct {
 	FilterDstPort  uint16
 	FilterPort     uint16
 	FilterTrackSkb bool
+	FilterPcap     string
```
brb, line 37: I'd suggest to drop all previous filters, and allow users to specify the filter pcap in the same way as with tcpdump. For example, `pwru --output-tuple 'host 1.1.1.1 and port 80'`. Let's do it as a follow-up.

brb (approved):
Very nice! Also, thanks a lot for code comments which made the changes easier to understand.

Some minor nits.

libpcap/inject.go
jschwinger233: Will revise, thanks for pointing out 👍

libpcap/compile.go
```diff
@@ -0,0 +1,202 @@
...
+#cgo LDFLAGS: -L/usr/local/lib -lpcap -static
+#include <stdlib.h>
+#include <pcap.h>
+*/
+import "C"
+
+type pcapBpfProgram C.struct_bpf_program
+
+const (
+	MaxBpfInstructions       = 4096
```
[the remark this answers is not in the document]
jschwinger233: You're right, but this is for compiling pcap-filter to cbpf, which is concise and I consider 4096 is sufficient so that we don't have to waste unnecessary memory.

For instance, we can implement equivalent filter to `pwru --filter-src-ip --filter-dst-ip --filter-src-port --filter-dst-port --filter proto --filter-tcp-flag syn` using only 19 pieces of cbpf instructions:

```
$ tcpdump -d 'src host 1.1.1.1 and dst host 2.2.2.2 and src port 23 and dst port 23 and tcp and tcp[tcpfl
ags] & tcp-syn != 0'
Warning: assuming Ethernet
(000) ldh      [12]
(001) jeq      #0x800           jt 2	jf 19
(002) ld       [26]
(003) jeq      #0x1010101       jt 4	jf 19
(004) ld       [30]
(005) jeq      #0x2020202       jt 6	jf 19
(006) ldb      [23]
(007) jeq      #0x84            jt 19	jf 8
(008) jeq      #0x6             jt 9	jf 19
(009) ldh      [20]
(010) jset     #0x1fff          jt 19	jf 11
(011) ldxb     4*([14]&0xf)
(012) ldh      [x + 14]
(013) jeq      #0x17            jt 14	jf 19
(014) ldh      [x + 16]
(015) jeq      #0x17            jt 16	jf 19
(016) ldb      [x + 27]
(017) jset     #0x2             jt 18	jf 19
(018) ret      #262144
(019) ret      #0
```

PR #202, opened 2023-06-28
main: use rlimit.RemoveMemlock()

lmb:
cilium/ebpf provides a helper which removes the memlock limit if necessary. Use it instead of unconditionally bumping RLIMIT_MEMLOCK.


tklauser (approved):
Nice, thanks.

PR #203, opened 2023-06-28
main: pass -target to bpf2go

lmb:
Use the -target flag to tell bpf2go which arch we are targeting. This cuts down on compiling useless targets (big endian x86 anyone?)

Also use $GOARCH from as set by go generate to reduce the necessary boilerplate.


lmb:
Should've mentioned that bpf2go does this for you: https://github.com/cilium/ebpf/blob/be9a38d9613995223b0479c853b1f89b611977af/cmd/bpf2go/main.go#L335-L337

Issue #205, opened 2023-07-01
kprobe_pwru.c:set_tuple() reads wrong L4 for ESP on some kprobes


jschwinger233:
## TL;DR

Current `set_tuple()` reads `l4_off` from `skb->transport_header` and then get `struct tcphdr *tcp` based on that:

https://github.com/cilium/pwru/blob/f470db0673c3756faaabab68a0e705b534f69482/bpf/kprobe_pwru.c#L205-L208

https://github.com/cilium/pwru/blob/f470db0673c3756faaabab68a0e705b534f69482/bpf/kprobe_pwru.c#L227-L228

But on some kprobes such as `xfrm_rcv_cb` and `xfrm6_rcv_cb`, `skb->head + skb->transport_header` points to the same position as `skb->head + skb->network_header`, causing wrong L4 tuple settings therefore.

## Long story

I was observing cilium with IPsec using pwru, while keeping curling from pod1@host1 to pod2@host2 using IPv6.

The pwru command below filters tcp syn traffic on host2 (receiving host):

```
$ s pwru --output-tuple --output-meta --filter-dst-port 80  --filter-dst-ip fd00:10:244:1::cf33 --filter-pcap 'ip6[40+13] & 0x3f = 0x2' --filter-netns 4026533408 
2023/07/01 12:43:37 Per cpu buffer size: 4096 bytes
2023/07/01 12:43:37 Attaching kprobes (via kprobe-multi)...
1498 / 1498 [-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00% ? p/s
2023/07/01 12:43:37 Attached (ignored 0)
2023/07/01 12:43:37 Listening for events..
               SKB    CPU          PROCESS                     FUNC
0xffff98ae4743eae8      0           [curl]              xfrm_rcv_cb netns=4026533408 mark=0xd00 ifindex=4 proto=dd86 mtu=1500 len=80 [fd00:10:244::834b]:24589->[fd00:10:244:1::cf33]:62723(tcp)
0xffff98ae4743eae8      0           [curl]             xfrm6_rcv_cb netns=4026533408 mark=0xd00 ifindex=4 proto=dd86 mtu=1500 len=80 [fd00:10:244::834b]:24589->[fd00:10:244:1::cf33]:62723(tcp)
[...]

PR #216, opened 2023-07-05
Derive the L4 hdr from the L3 hdr

jschwinger233:
For some kernel functions like xfrm6_rcv_cb, skb->transport_header is wrongly set to the same value as skb->network_header, causing confusing output showing unexpected tcp port.

This commit lets kprobe_pwru.c:set_tuple() calculate L4 header based on L3 header. The code to calculate ipv6 header length is simply revised from
https://github.com/cilium/cilium/blob/e3a8cfe4ac56883b55cb9f1b8fdfa3ac7f54aaf6/bpf/lib/ipv6.h#L81

Fixes: #205

### DIFF
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -14,6 +14,7 @@
 #include "bpf/bpf_helpers.h"
 #include "bpf/bpf_core_read.h"
 #include "bpf/bpf_tracing.h"
+#include "bpf/bpf_ipv6.h"
 
 #define PRINT_SKB_STR_SIZE    2048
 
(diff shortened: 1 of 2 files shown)


bpf/kprobe_pwru.c
```diff
@@ -206,12 +195,15 @@ set_tuple(struct sk_buff *skb, struct tuple *tpl) {
 		BPF_CORE_READ_INTO(&tpl->daddr, ip4, daddr);
 		tpl->l4_proto = BPF_CORE_READ(ip4, protocol);
 		tpl->l3_proto = ETH_P_IP;
+		l4_off = l3_off + BPF_CORE_READ_BITFIELD_PROBED(ip4, ihl) * 4;
```
brb, line 198: I assume that `ihl` includes IPv4 options, right?

brb: The internet says that yes, it's included :+1:

brb (approved):
Thanks!

PR #223, opened 2023-07-08
Fix code not matching the comment

mozillazg:
https://github.com/cilium/pwru/pull/207

cc @jschwinger233

### DIFF
--- a/internal/libpcap/inject.go
+++ b/internal/libpcap/inject.go
@@ -162,7 +162,7 @@ func InjectFilter(program *ebpf.ProgramSpec, filterExpr string) (err error) {
 		// stack area is safe to use. Here we use stack from -40
 		// because -32, -24, -16 are reserved for pcap-filter ebpf, see
 		// the comments in compile.go
-		StackOffset: 40,
+		StackOffset: 32,
 	})
 	if err != nil {
 		return


internal/libpcap/inject.go
```diff
@@ -162,7 +162,7 @@ func InjectFilter(program *ebpf.ProgramSpec, filterExpr string) (err error) {
 		// stack area is safe to use. Here we use stack from -40
```
mozillazg, line 162: Should I update the comment instead?
```suggestion
		// stack area is safe to use. Here we use stack from -48
```

internal/libpcap/inject.go
```diff
@@ -162,7 +162,7 @@ func InjectFilter(program *ebpf.ProgramSpec, filterExpr string) (err error) {
...
 		// because -32, -24, -16 are reserved for pcap-filter ebpf, see
 		// the comments in compile.go
-		StackOffset: 40,
+		StackOffset: 32,
```
jschwinger233, line 165: I think you're right.

`StackOffset: 32` means cloudfare/cbpfc is able to use stack area from r10-32 downwards, in another word, the first available stack pointer points to r10-40.  Your change made it in line with the real situation, thanks.

jschwinger233 (approved):

internal/libpcap/inject.go
jschwinger233: Let's change the offset instead.

PR #226, opened 2023-07-14
Update demo cast

brb:
And some minor changes.

Just in case - [pwru.cast.gz](https://github.com/cilium/pwru/files/12050505/pwru.cast.gz)


jschwinger233 (approved):
Perfect!  :+1:

PR #242, opened 2023-08-28
Raise rlimit nofile to 8192

jschwinger233:
Previous limit 4096 isn't enough for `--all-kmods`, lsof shows pwru uses as many as 5269 fds in that case.

Siged-off-by: Zhichuan Liang <gray.liang@isovalent.com>

### DIFF
--- a/main.go
+++ b/main.go
@@ -36,8 +36,8 @@ func main() {
 	}
 
 	if err := unix.Setrlimit(unix.RLIMIT_NOFILE, &unix.Rlimit{
-		Cur: 4096,
-		Max: 4096,
+		Cur: 8192,
+		Max: 8192,
 	}); err != nil {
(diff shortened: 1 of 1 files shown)


jschwinger233:
CI checks failed due to `failed to set temporary rlimit: operation not permitted` ...

main.go
```diff
@@ -36,8 +36,8 @@ func main() {
 	}
 
 	if err := unix.Setrlimit(unix.RLIMIT_NOFILE, &unix.Rlimit{
-		Cur: 4096,
-		Max: 4096,
+		Cur: unix.RLIM_INFINITY,
```
brb, line 39: Maybe let's set to 2x4096? This would help us to avoid the CI fixing :pizza:

brb (approved):

PR #251, opened 2023-09-08
Allow example to run as script

azzid:
Doing `apply`, `logs`, `delete` in sequence as fast as a script does has the unfortuante effect that it does logs before the pod is ready.

Adding a `wait` sorts that.

Sending `ctrl+c` to break `logs` will however terminate `bash` before `delete` can run.

Putting `delete` in a `trap` sorts that.


brb (approved):
Thanks!

PR #252, opened 2023-09-19
Use bpf2bpf to simplify injection

jschwinger233:
By using bpf2bpf:

1. We don't have to search for registers holding skb->data and skb-data_end, because they must be at R4 and R5 as arguments.
2. We no longer need a bpf_printk() to be replaced, so pwru developers can use printk for debugging again.
3. We don't need to assume clang behavior or rely on specific clang version.
4. We can delete a lot of code to simplify logic.


brb (approved):
Amazing!

PR #257, opened 2023-09-26
Add --filter-ifname

jschwinger233:
Set this flag to filter skb from/to the specific interface under the specific netns. If flag --filter-netns is not set, pwru uses current netns.

### DIFF
--- a/internal/pwru/types.go
+++ b/internal/pwru/types.go
@@ -25,10 +25,11 @@ type Flags struct {
 
 	KernelBTF string
 
-	FilterNetns    uint32
+	FilterNetns    string
 	FilterMark     uint32
 	FilterFunc     string
 	FilterTrackSkb bool
+	FilterIfname   string
 	FilterPcap     string
 
 	OutputTS         string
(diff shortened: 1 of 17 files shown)


brb:
@jschwinger233 Maybe let's resolve ifname to ifindex from the user space, and then use ifindex in the kprobe?

jschwinger233:
@brb That was my idea, but I ended up giving it up because I didn't know how to switch netns using ID.

I know how to switch netns: https://www.weave.works/blog/linux-namespaces-golang-followup, I don't know how to get NsHandle from ID.

I looked up [vishvananda/netns](https://pkg.go.dev/github.com/vishvananda/netns), there are 5 functions to get a NsHandle: `GetFromDocker`, `GetFromName`, `GetFromPath`, `GetFromPid`, `GetFromThread`, but none accepts a netns ID.

jschwinger233:
I think NsHandle is a wrapper of fd:
https://github.com/vishvananda/netns/blob/16c2fa0b2f57726d0e9e4acd228a63c322954f31/netns_linux.go#L101-L106

The value should be a small integer like 5, 6, 7, instead of something like 4026531834

brb:
Hmm, you are right. And on Linux getting FD by inode is not possible :-/ Maybe we should change the semantics of `filter-netns`, and require it to point to a netns path? That would also improve the UX.

jschwinger233:
I agree that netnsPath improves UX, but `netns` field in output of `pwru --output-meta` shows inode. If someone wants to filter a specific netns based on previous pwru output, it would be inconvenient.

(For me, I can use `lsns` to search pid by inode, but it's still an extra burden for users.)

Speaking of output fields, one more thing is ifindex field in output of `pwru --output-meta`, I'd love to convert that ifindex to ifname and show something like `if=3(eth0)`, still it's hard to parse ifindex in userspace. We can achieve it in bpf program by fetching `skb->dev->name`, not sure if we want to do so.

brb:
> I agree that netnsPath improves UX, but netns field in output of pwru --output-meta shows inode. If someone wants to filter a specific netns based on previous pwru output, it would be inconvenient.

My main worry with the current ifname approach is that in the critical path we have a slow operation (string vs int cmp).

WDYT about having `--filter-netns` (by path) and `--filter-netns-by-inode`, and then making `--filter-netns-by-inode` mutual exclusive with `--filter-ifname` (which we convert to ifindex in the user space)?

brb:
> I'd love to convert that ifindex to ifname and show something like if=3(eth0), still it's hard to parse ifindex in userspace.

It's been on my list since forever :sweat_smile:  When the prog starts, we could fetch ifindices from all netns, and then build a cache (prone to races, but it's fine) - `key=[netns_inode, ifindex] => val=iface_name`.

brb:
> So how about letting --filter-netns accept multiple forms of arguments:

SGTM! Just maybe let's make path a default.

internal/pwru/types.go
```diff
@@ -53,15 +54,16 @@ func (f *Flags) SetFlags() {
 	flag.StringSliceVar(&f.KMods, "kmods", nil, "list of kernel modules names to attach to")
 	flag.BoolVar(&f.AllKMods, "all-kmods", false, "attach to all available kernel modules")
 	flag.StringVar(&f.FilterFunc, "filter-func", "", "filter kernel functions to be probed by name (exact match, supports RE2 regular expression)")
-	flag.Uint32Var(&f.FilterNetns, "filter-netns", 0, "filter netns inode")
+	flag.StringVar(&f.FilterNetns, "filter-netns", "", "filter netns (\"proc/<pid>/ns/net\", \"inode:<inode>\")")
```
brb, line 57: Nit: s/proc/\/proc/

brb:
@jschwinger233 Very nice! One nit, and then we can :shipit:

brb:
@jschwinger233 Could you rebase to resolve the conflict?

brb (approved):
:rocket:

PR #259, opened 2023-09-28
Output iface name when --output-meta is set

brb:
This commits makes --output-meta to output not only ifindex, but also
iface name.

To achieve that, we first build a netns_inode => ifindex => iface_name
cache. This is done by entering each /proc/$PID/ns/net, and then quering
ifaces. Then, when a pwru process receives an BPF event, it queries the
cache (netns inode and ifindex are set in the event).

Example:

    ... [curl]  nf_hook_slow netns=4026531840 mark=0x0 iface=3(wlan0)


internal/pwru/output.go
```diff
@@ -212,3 +240,99 @@ func getKFreeSKBReasons(spec *btf.Spec) (map[uint64]string, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	ifaces := make(map[uint32]string)
+	for _, link := range msg {
+		ifaces[link.Index] = link.Attributes.Name
+	}
+
+	runtime.UnlockOSThread()
```
jschwinger233, line 335: Duplicate? 311L has a `defer runtime.UnlockOSThread()`

brb: Oops :+1:

PR #263, opened 2023-09-29
Report skb->protocol in host byte order

tklauser:
Currently, the proto field (corresponding to skb->protocol which is in network byte order) is reported in pwru's output in network byte order. Report it in host byte order instead so it can easily be matched to the ethertype.

Reported by @darox ref. https://cilium.slack.com/archives/C0300URJQM7/p1695985654194089


brb (approved):
Thanks!

PR #266, opened 2023-10-14
Replace objs with collection

Asphaltt:
Go is a static programming language which is lack of dynamic struct for our `print_skb_map` case.

But the go-ebpf library provides a dynamic way for this case, which is `NewCollectionWithOptions()`. With this way, it is unnecessary to define type interfaces and their getters.


brb (approved):
Thanks!

PR #267, opened 2023-10-16
Readme: Improve instructions

darox:
This commit improves the instructions for building pwru by listing additional tools that are required to build it. Besides that it adds a step for building the libpcap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,16 +141,9 @@ See [docs/vagrant.md](docs/vagrant.md)
 
 ### Building
 
-
-Build libpcap before building pwru:
 ```
-make libpcap/libpcap.a
+make
 ```
-
-Now you can build pwru:
-```
-make 
-``````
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -136,12 +136,21 @@ See [docs/vagrant.md](docs/vagrant.md)
 
 * Go >= 1.16
 * LLVM/clang >= 1.12
+* Bison 
+* Lex/Flex >= 2.5.31
 
 ### Building
 
+
+Build libpcap before building pwru:
```
brb, line 145: I think this is not needed, as the dependency https://github.com/cilium/pwru/blob/main/Makefile#L17 should trigger the libpcap build process.

darox: You're right.

brb (approved):
Thanks!

PR #268, opened 2023-10-16
Makefile: Add help target

darox:
This commit adds a help target to the Makefile. I initially also wanted to change the default target to help, but this breaks the current workflow of users. If wished I can still do that.

image


brb:
Thanks, that's a nice addition! Could we add the `make pwru` target to the list (the default)?

darox:
> Thanks, that's a nice addition! Could we add the `make pwru` target to the list (the default)?

Done `pwru` is now the default target:
image

brb (approved):
Thanks!

PR #280, opened 2023-10-27
Find process using pid instead of tid

jschwinger233:
Previously we use tid to find process, leading to misleading output under some situations. 

According to documentation, `bpf_get_current_pid_tgid` returns `current->tgid << 32 | current->pid`. Kernel's view of the PID in user space is usually presented as the thread ID, and kernel's tgid in user space is seen as PID.


brb (approved):
Thanks!

PR #289, opened 2023-11-22
Makefile clean-up

qmonnet:
Minor clean-ups from taking a quick look at the Makefile:

- Makefile: Fix awk command in "make help"
- Makefile: Remove compiler versions for "make local-release"


brb (approved):
Thanks!

PR #290, opened 2023-11-22
internal/pwru: Implement a proper help function

qmonnet:
With `./pwru --help` or `./pwru -h`, pwru:

  1. displays the help message,
  2. prints `pflag: help requested`,
  3. exits with error code 2.

Steps 2 and 3 are not expected, the interactive help should return 0 without displaying an error message. This is because of how `pflags.ParseAll()` works. From [its documentation][doc]: _The return value will be ErrHelp if -help was set but not defined._

Let's implement a dedicated function for displaying the help message (reusing `pflags.Usage()` and what we had in pwru already), and call it when users pass `--help` or `-h` to avoid this error and non-zero return value.

[doc]: https://pkg.go.dev/github.com/spf13/pflag#FlagSet.ParseAll


brb (approved):
Thanks!

PR #294, opened 2023-12-06
Fix get_func_ip for --filter-trace-tc

### LINKED ISSUE #293: option --filter-trace-tc not support
Hello guys,

I want to use pwru to trace cilium bpf program execution.

But when I run pwru by below command:

```bash
./pwru --filter-trace-tc "tcp and host 10.0.1.90 and port 80"  --output-tuple  --output-file tcp2
```

but got below error:

```bash
......
        ; event.addr = bpf_get_func_ip(ctx);
        26: (bf) r1 = r6
        27: R0=invP1 R1_w=ctx(id=0,off=0,imm=0) R6=ctx(id=0,off=0,imm=0) R7=ptr_sk_buff(id=0,off=0,imm=0) R10=fp0 fp-8=????mmmm fp-16=00000000 fp-24=00000000 fp-32=0mm0mmmm fp-40=mmmmmmmm fp-48=mmmmmmmm fp-56=mmmmmmmm fp-64=mmmmmmmm fp-72=00000000 fp-80=00000000 fp-88=00000000 fp-96=00000000 fp-104=mmmmmmmm fp-112_w=ptr_ fp-120=00000000 fp-128=0000mmmm
        27: (85) call bpf_get_func_ip#173
        func bpf_get_func_ip#173 not supported for program type 3
        processed 437 insns (limit 1000000) max_states_per_insn 1 total_states 28 peak_states 28 mark_read 28

program fentry_tc: load program: operation not supported: func bpf_get_func_ip#173 not supported for program type 3 (1020 line(s) omitted)
```

If not use `--filter-trace-tc` option, it works well.

Asphaltt:
Fix #293 .

`bpf_get_func_ip()` is supported since 5.15 kernel.

- [bpf: Add bpf_get_func_ip helper for tracing programs](https://github.com/torvalds/linux/commit/9b99edcae5c80c8fb9f8e7149bae528c9e610a72)

As a result, we have to figure out a way to avoid `bpf_get_func_ip()`.

Then, I realise that something in `/proc/kallsyms` may help.

```bash
ffffffffc0024020 t bpf_prog_a04f5eef06a7f555_dummy	[bpf]
ffffffffc0024094 t bpf_prog_a04f5eef06a7f555_dummy	[bpf]
ffffffffc002ac4c t bpf_prog_0b3a2ce7164b50f3_fentry_tc	[bpf]
ffffffffc002add0 t bpf_prog_0ca2b1941f84fe9b_fexit_tc	[bpf]
```

Next, read the kernel source code.

- [bpf_prog_ksym_set_name()](https://github.com/torvalds/linux/blob/v5.8/kernel/bpf/core.c#L539)

So, the bpf prog info in `/proc/kallsyms` maybe formats with `ADDR t bpf_prog_<tag>_<name>\t[bpf]`.

Finally, I figure out the solution:

1. Record `tag` to `ADDR` in a map.
2. Get `tag` from bpf prog info.
3. Write `tag`'s `ADDR` to bpf code.

Great! And optimise the outputing name of bpf prog BTW.

### DIFF
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -15,6 +15,8 @@
 
 const static bool TRUE = true;
 
+volatile const static __u64 BPF_PROG_ADDR = 0;
+
 union addr {
 	u32 v4addr;
 	struct {
@@ -372,7 +374,7 @@ int BPF_PROG(fentry_tc, struct sk_buff *skb) {
 		return BPF_OK;
 
 	event.skb_addr = (u64) skb;
-	event.addr = bpf_get_func_ip(ctx);
+	event.addr = BPF_PROG_ADDR;
 	bpf_map_push_elem(&events, &event, BPF_EXIST);
 
 	return BPF_OK;
--- a/internal/pwru/bpf_prog.go
+++ b/internal/pwru/bpf_prog.go
@@ -14,6 +14,8 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+type BpfProgName2Addr map[string]uint64
+
 func listBpfProgs(typ ebpf.ProgramType) ([]*ebpf.Program, error) {
 	var (
 		id  ebpf.ProgramID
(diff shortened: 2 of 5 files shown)


jschwinger233 (commented):
bpf prog tag is not unique, what if two different prog (with different addresses) have the same tag?

Asphaltt:
> bpf prog tag is not unique

Yes, the tag is not unique. [bpf_prog_calc_tag](https://github.com/torvalds/linux/blob/v6.6/kernel/bpf/core.c#L280)

How about name -> addr instead of tag -> addr?

I think, it's to map addr -> name for bpf prog actually, no matter whether tag is unique.

jschwinger233:
>  it's to map addr -> name for bpf prog actually, no matter whether tag is unique.

You are right, after all pwru will translate address to symbol name, so it doesn't have to be the precise address as long as the address is mapped to correct address.

brb:
@Asphaltt :wave: Could you rebase your PR against the latest main branch? Also, does it still need a review?

Asphaltt:
Yes, it's ready to review again.

jschwinger233 (approved):
I'll approve it now, but want to bring this up: the `event.addr` may hold the wrong rip value, because two addresses could have one same symbol. As discussed before, that doesn't affect the symbol resolution because even a "wrong" rip can be matched to the right symbol. However, this could make a difference when we want to distinguish which exact bpf prog is called.

brb:
> I'll approve it now, but want to bring this up: the event.addr may hold the wrong rip value, because two addresses could have one same symbol. As discussed before, that doesn't affect the symbol resolution because even a "wrong" rip can be matched to the right symbol. However, this could make a difference when we want to distinguish which exact bpf prog is called.

@Asphaltt maybe you could add this as a code comment?

Asphaltt:
Sure.

PR #305, opened 2023-12-22
Bump Go to 1.21.5 and check release build in CI

tklauser:
See commits


jschwinger233 (approved):
Thanks! BTW CI test runs `make release` on PR: https://github.com/cilium/pwru/blob/main/.github/workflows/test.yml#L30

tklauser:
> Thanks! BTW CI test runs `make release` on PR: https://github.com/cilium/pwru/blob/main/.github/workflows/test.yml#L30

TIL, thanks. I guess we can drop the second commit running the release workflow on every PR.

PR #308, opened 2024-01-10
Remove unused relative timestamp config option

tklauser:
The option setting it was superseded by --timestamp=relative and is thus unused since commit 93fee0ade78d ("Make timestamp optional").


jschwinger233 (approved):
Thanks!

PR #309, opened 2024-01-11
Let pwru work without /sys/kernel/debug/tracing/available_filter_functions

jschwinger233:
/sys/kernel/debug/tracing/available_filter_functions is crucial for kprobe-multi but not necessary for kprobe, while pwru refuses to work without it.

This PR let pwru fallback to kprobe backend if /sys/kernel/debug/tracing/available_filter_functions is not available.


tklauser (approved):
Thanks!

PR #323, opened 2024-02-14
gh/workflows: misc test changes

brb:
See commit msgs.


jschwinger233 (approved):
Thanks! Maybe pwru-test action under .github/actions/pwru-test/ could use the newer version of lvh too.

PR #349, opened 2024-04-07
reduce the build context sent during a docker build

lpmi-13:
Even though these files don't end up in the final image, they're unnecessary for the build image and don't need to be sent with the build context.


brb (approved):
Thanks!

PR #350, opened 2024-04-08
Avoid printing garvage skb written by previous writer

YutaroHayakawa:
After we consume all of the print_skb_map slot, we start to reuse the slot. On the reused buffer, there's a previous skb string and we don't clear that. So, when the new skb string is shorter than the previous one, we print the garvage string. Keep the length of the string in the map and truncate unnecessary part from the output to avoid it.

### DIFF
--- a/internal/pwru/output.go
+++ b/internal/pwru/output.go
@@ -5,6 +5,7 @@
 package pwru
 
 import (
+	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


internal/pwru/output.go
```diff
@@ -247,10 +248,20 @@ func getStackData(event *Event, o *output) (stackData string) {
...
+
+	length := binary.NativeEndian.Uint32(b[:4])
+
+	// Bounds check
+	if int(length+4) > len(b) {
+		return ""
 	}
-	return skbData
+
+	return "\n" + string(b[4:4+length])
```
brb, line 264: Didn't you mean `return string(b[4:4+length]) + "\n"` instead?

YutaroHayakawa: No, this is intended. Currently, we start to print skb right after the function name

```
0xffff88886f430d00      6 [udp_main:913171]             kfree_skbmem(struct sk_buff){
 (union){
  (struct){
   (union){
    .dev = (struct net_device *)0x000000000e329f1c,
    .dev_scratch = (long unsigned int)18446612686734655488,
   },
```

This makes it a bit cleaner.

```
0xffff88886f430d00      6 [udp_main:913171]             kfree_skbmem
(struct sk_buff){
 (union){
  (struct){
   (union){
    .dev = (struct net_device *)0x000000000e329f1c,
    .dev_scratch = (long unsigned int)18446612686734655488,
   },
```

brb (approved):
Thanks!

PR #355, opened 2024-04-19
Implement --filter-non-skb-funcs

jschwinger233:
By enabling --filter-track-skb-by-stackid, we can track kernel functions without skb parameter. 

For example, to observe xfrm state lookup, we can use command: `pwru --filter-non-skb-funcs xfrm_state_look_at,xfrm_state_lookup,xfrm_state_lookup_byaddr,xfrm_state_lookup_byspi`. 

Also, this is another step towards "tailcall tracing". This PR allows us to attach bpf helpers (that are probably not skb functions). By fetching bpf helpers' caller pc and converting pc into symbol, we'll see the bpf prog name.

### DIFF
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -78,7 +78,7 @@ struct {
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
-	__type(key, struct sk_buff *);;
+	__type(key, struct sk_buff *);
 	__type(value, __u64);
 	__uint(max_entries, MAX_TRACK_SIZE);
 } skb_stackid SEC(".maps");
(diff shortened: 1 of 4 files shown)


bpf/kprobe_pwru.c
```diff
@@ -393,6 +393,17 @@ PWRU_ADD_KPROBE(3)
 PWRU_ADD_KPROBE(4)
 PWRU_ADD_KPROBE(5)
 
+SEC("kprobe/skb_by_stackid")
+int kprobe_skb_by_stackid(struct pt_regs *ctx) {
+	u64 stackid = get_stackid(ctx);
+
+	struct sk_buff **skb = bpf_map_lookup_elem(&stackid_skb, &stackid);
+	if (skb && *skb)
+		return kprobe_skb(*skb, ctx, PWRU_HAS_GET_FUNC_IP);
```
jschwinger233, line 402: I should pass the `stackid` as an arguement of `kprobe_skb` so we don't have to get_stackid() for twice.

bpf/kprobe_pwru.c
```diff
@@ -393,6 +391,17 @@ PWRU_ADD_KPROBE(3)
 PWRU_ADD_KPROBE(4)
 PWRU_ADD_KPROBE(5)
 
+SEC("kprobe/skb_by_stackid")
+int kprobe_skb_by_stackid(struct pt_regs *ctx) {
+	u64 stackid = get_stackid(ctx);
+
+	struct sk_buff **skb = bpf_map_lookup_elem(&stackid_skb, &stackid);
+	if (skb && *skb)
```
brb, line 399: Why `if (skb)` is not enough?

jschwinger233: You are right, it's MS Copilot! Let me fix it.

jschwinger233: Okay.. you merged ...

PR #383, opened 2024-05-30
fix: Get bpf prog entry func name from insns

Asphaltt:
It is possible to get wrong entry func name of bpf prog when there is info of multiple bpf progs in the BTF.

Instead, get the very first symbol from bpf prog's instructions as its entry func name.

### DIFF
--- a/internal/pwru/bpf_prog.go
+++ b/internal/pwru/bpf_prog.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 
 	"github.com/cilium/ebpf"
-	"github.com/cilium/ebpf/btf"
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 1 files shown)


jschwinger233 (approved):
In the first place I had some doubts in this symbol approach. For example, `bpftrace -e 'tracepoint:syscalls:sys_enter_epoll_wait {printf("");}'` loads a prog which `bpftool p d x` shows no symbol, where this PR might not work. But some local tests convinced me. LGTM.

Asphaltt:
> In the first place I had some doubts in this symbol approach.

What bpf progs we can trace are those have BTF info. When bpf prog has BTF info, it is able to extract the symbol from its instructions, as its BTF info contains the line info and function info.

Ref:  [Instructions() source code](https://github.com/cilium/ebpf/blob/b8dc0ee25417ce7cd4a6feb48be42c0615ee9043/info.go#L296)

brb (approved):
Thanks!

PR #391, opened 2024-06-29
--filter-track-skb can track skbs re-built from veth_convert_skb_to_xdp_buff

jschwinger233:
When XDP is attached to a veth, skbs will be consumed and re-created on that veth. This is done in the function veth_convert_skb_to_xdp_buff():

```
// drivers/net/veth.c
static int veth_convert_skb_to_xdp_buff(struct veth_rq *rq,
					struct xdp_buff *xdp,
					struct sk_buff **pskb)
{
	struct sk_buff *skb = *pskb;
[...]
		nskb = build_skb(page_address(page), PAGE_SIZE);
[...]
		skb_copy_header(nskb, skb);
[...]
		consume_skb(skb);
		skb = nskb;
[...]
}
```

This causes problems for pwru --filter-track-skb because of the new skb addresses. I ran into a lot of situations where I lost track of NAT-ed traffic at veth when cilium kind cluster is created by "kind.sh --xdp".

This patch allows pwru to keep track of the new skbs at XDP-attached veth devices.

### DIFF
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -88,6 +88,13 @@ struct {
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
+	__type(key, __u64); // pid_tgid
+	__type(value, __u64); // struct sk_buff **
+	__uint(max_entries, MAX_TRACK_SIZE);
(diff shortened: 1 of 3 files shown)


brb (approved):
Thanks!

PR #410, opened 2024-08-06
Fix tracking skb

jschwinger233:
This PR uses &skb to track skb instead of skb->head to prevent several issues. XDP tracing is also taken care of by introducing a new bpf map `xdp_dhs_skb_heads`.


tklauser:
Requesting an additional review from @brb since I'm not that familiar with the BPF bits of pwru.

tklauser (approved):

brb (approved):
:rocket:

Issue #409, opened 2024-08-06
`--output-skb` not working on eks on 5.10 kernel


jibi:
```
# ./pwru --output-skb
2024/08/06 08:05:10 Failed to load objects: Verifier error: load program: invalid argument:
	func#0 @0
	func#1 @59
	func#2 @1133
	func#3 @1249
	BPF_STX uses reserved fields
	processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0

program kprobe_skb_1: load program: invalid argument: BPF_STX uses reserved fields (5 line(s) omitted)
```

things I tried so far (no luck):
* building it with different versions of clang (12, 17, 18)
* building the bpf programs with `-O2`

jibi:
getting rid of atomic operations seem to fix the issue:

```diff
diff --git a/bpf/kprobe_pwru.c b/bpf/kprobe_pwru.c
index 092f53a..0ad2c9e 100644
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -323,7 +323,8 @@ set_skb_btf(struct sk_buff *skb, typeof(print_skb_id) *event_id) {

        p.type_id = bpf_core_type_id_kernel(struct sk_buff);
        p.ptr = skb;
-       id = __sync_fetch_and_add(&print_skb_id, 1) % 256;
+       //id = __sync_fetch_and_add(&print_skb_id, 1) % 256;
+       id = (print_skb_id  + 1) % 256;

        v = bpf_map_lookup_elem(&print_skb_map, (u32 *) &id);
        if (!v) {
@@ -366,7 +367,8 @@ set_shinfo_btf(struct sk_buff *skb, typeof(print_shinfo_id) *event_id) {
        p.type_id = bpf_core_type_id_kernel(struct skb_shared_info);
        p.ptr = shinfo;

-       id = __sync_fetch_and_add(&print_shinfo_id, 1) % 256;
+       //id = __sync_fetch_and_add(&print_shinfo_id, 1) % 256;
+       id = (print_shinfo_id  + 1) % 256;

        v = bpf_map_lookup_elem(&print_shinfo_map, (u32 *) &id);
        if (!v) {
```

PR #412, opened 2024-08-07
bpf: Use percpu array to replace __sync_fetch_and_add

jschwinger233:
Fixes: https://github.com/cilium/pwru/issues/409


brb (approved):
Thanks!

PR #424, opened 2024-09-24
Fix broken --filter-trace-tc and --filter-trace-xdp

Asphaltt:
When I want to troubleshoot a network issue with `--filter-trace-tc`, it cannot to get the address of bpf prog.

After fixing `--filter-trace-tc` issue, I find it fails to run with `--filter-trace-xdp` too. So, I fix the broken `--filter-trace-xdp` BTW.

Signed-off-by: Leon Hwang <hffilwlqm@gmail.com>

### DIFF
--- a/internal/pwru/ksym.go
+++ b/internal/pwru/ksym.go
@@ -58,7 +58,7 @@ func ParseKallsyms(funcs Funcs, all bool) (Addr2Name, BpfProgName2Addr, error) {
 	scanner := bufio.NewScanner(file)
 	for scanner.Scan() {
 		line := strings.Split(scanner.Text(), " ")
-		name, isBpfProg := extractBpfProgName(line[2])
+		name := strings.ReplaceAll(line[2], "\t", "")
 		if all || (funcs[name] > 0) {
 			addr, err := strconv.ParseUint(line[0], 16, 64)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


internal/pwru/ksym.go
```diff
@@ -57,8 +57,12 @@ func ParseKallsyms(funcs Funcs, all bool) (Addr2Name, BpfProgName2Addr, error) {
...
 	scanner := bufio.NewScanner(file)
 	for scanner.Scan() {
-		line := strings.Split(scanner.Text(), " ")
-		name, isBpfProg := extractBpfProgName(line[2])
+		line := strings.Fields(scanner.Text())
+		isBpfProg := len(line) > 3 && line[3] == "[bpf]"
+		name := line[2]
+		if isBpfProg && len(name) > 26 {
+			name = name[26:]
+		}
```
jschwinger233, line 65: This breaks --filter-track-bpf-helpers because pwru relies on  `[bpf]` suffix to collect bpf functions.

jschwinger233 (commented):
--track-bpf-helpers prefers the raw ksym such as `bpf_prog_71868b7d3ecebc37_tail_ipv6_ct_ingress[bpf]` that clearly tells it's a bpf prog. The tag shows which exact bpf prog is triggered, considered useful during cilium upgrade when there are multiple bpf prog with a same name but different tags.

That's why `ParseKallsyms` was changed to use full ksym without stripping any parts.

Do you think we can make --trace-tc and --trace-xdp also use the raw ksym? For me it's also valuable to show tag in case there are bpf with duplicate names.

Also asking @brb 's opinion.

Asphaltt:
> to show tag in case there are xdp/tc bpf with duplicate names.

Aha, previously, `--filter-trace-{tc,xdp}` does not care about the tag, and focuses on prog name only.

However, it's acceptable to show the raw ksym including tag for `--filter-trace-{tc,xdp}`.

I'll update the commit later.

jschwinger233 (approved):
:+1:

PR #430, opened 2024-10-05
Print drop reason from sk_skb_reason_drop. Catch up with kernel 6.11 changes

DanPartelly:
This commit extracts a drop reason from sk_skb_reason_drop(), and prints it when sk_skb_drop_reason() is called.

This function was introduced in kernel v6.11-rc1. At the same time, kfree_skb_reason()   became an inline helper which just calls  sk_skb_reason_drop().
See: https://github.com/torvalds/linux/commit/ba8de796baf4bdc03530774fb284fe3c97875566

Summary of changes:
- bpf: teach kprobe_skb() to capture the 3rd function argument in the event map
-  go: teach getKFreeSKBReasons() to cope with changes to kfree_skb_reason and introduction of sk_skb_reason_drop() in kernel 
-  go: teach getOutFuncName() about sk_skb_reason_drop()

### DIFF
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -86,6 +86,7 @@ struct event_t {
 	struct tuple tuple;
 	s64 print_stack_id;
 	u64 param_second;
+	u64 param_third;
 	u32 cpu_id;
 } __attribute__((packed));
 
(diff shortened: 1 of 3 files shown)


brb (approved):
Thanks!

PR #440, opened 2024-10-21
gh/workflows: Bump LVH images

brb:


.github/workflows/test.yml
```diff
@@ -43,7 +43,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kernel: [ '5.4-20240201.165956', '5.10-20240201.165956', '5.15-20240201.165956', '6.1-20240201.165956', 'bpf-next-20240204.012837' ]
+        kernel: [ '5.4-20241218.004849', '5.10-20241218.004849', '5.15-20241218.004849', '6.1-20241218.004849', '6.6-20241218.004849', 'bpf-next-20250105.013256' ]
```
tklauser, line 46: Any worth in also adding 6.12 kernel or was it omitted on purpose?

brb: Good question. AFAIK (from https://www.kernel.org/) the 6.12 is not LTS?

tklauser: According to https://www.kernel.org/category/releases.html 6.12 will be an LTS release.

brb: :+1: added, ty!

tklauser (approved):

PR #444, opened 2024-10-22
README.md: Update Clang/LLVM dependency to 12

joestringer:
Relying on Clang 1.12 doesn't make sense. Update it to 12.


brb (approved):
:sweat_smile:

Issue #262, opened 2023-09-29
Multi-arch container image


lizrice:
The container image is currently x86 only, would be great to have a multi-arch image so people can easily run as a container on ARM platforms as well

lizrice:
This is not a full solution but maybe points in a useful direction: I was able to make some small changes to the Dockerfile to get a multi-arch image that works on my ARM Mac at least. 

```
FROM --platform=$TARGETPLATFORM ubuntu:22.04 AS build
ARG TARGETPLATFORM
ARG BUILDPLATFORM
ARG TARGETARCH
ARG BUILDARCH
RUN echo "I am running on $BUILDPLATFORM, building for $TARGETPLATFORM" > /log

ENV PATH $PATH:/usr/local/go/bin

RUN apt update -y -q && \
    DEBIAN_FRONTEND=noninteractive apt install --no-install-recommends -y -q \
        curl \
        build-essential \
        ca-certificates \
        wget \
        gnupg2 \
        git \
        llvm \
        clang \
        gcc flex bison gcc-aarch64* libc6-dev-arm64-cross
RUN curl -s https://storage.googleapis.com/golang/go1.20.5.linux-$TARGETARCH.tar.gz | tar -v -C /usr/local -xz

WORKDIR /pwru
COPY . .
RUN TARGET_GOARCH=${TARGETARCH} GOARCH=${BUILDARCH} make 
RUN chmod a+x /pwru

FROM --platform=$TARGETPLATFORM busybox
COPY --from=build /pwru/pwru /usr/local/bin/
```
Using a [Docker container driver](https://docs.docker.com/build/drivers/docker-container/) I can locally build and push a multi-arch image like this: `docker buildx build --platform linux/amd64,linux/arm64 --tag <owner>/pwru --push  .`  (kinda slowly due to QEMU). But I couldn't quickly see a script or action or something that could be updated to run this build

mstrYoda:
Are there any progress on here? Would love to see official arm64 image.

brb:
Thanks for asking, but unfortunately no progress. We would love to see a contribution for it.

brb:
We still need to push the arm64 images to the Docker registry.

PR #449, opened 2024-11-01
Dockerfile: build arm64 variants

akerouanton:
- Closes https://github.com/cilium/pwru/issues/262

Make it possible to build both amd64 and arm64 Docker images. This can be done with:

    docker build --platform linux/amd64,linux/arm64 --tag pwru  .

GHA release workflow only handles GH release artifacts, so the out-of-tree CI job or manual process used to publish images on Docker Hub will have to be updated accordingly.

Also, fix the release script used during image build to support cross-compilation of amd64 on arm64 machines.

(Note that a standalone Docker Engine has to be configured with the containerd snapshotter enabled to build multi-arch images. See https://docs.docker.com/engine/storage/containerd/)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,17 @@
-FROM docker.io/library/golang:1.23.1 AS build
+FROM --platform=$BUILDPLATFORM docker.io/library/golang:1.23.1 AS build
+ARG TARGETPLATFORM
+ARG BUILDPLATFORM
+ARG TARGETARCH
 
-RUN apt update && \
-    apt install -y make git clang-15 llvm curl gcc flex bison gcc-aarch64* libc6-dev-arm64-cross && \
+RUN gcc_pkg=$(if [ "${TARGETARCH}" = "arm64" ]; then echo "aarch64"; else echo "x86-64"; fi)  && \
+    apt update && \
+    apt install -y make git clang-15 llvm curl gcc flex bison gcc-${gcc_pkg}* libc6-dev-${TARGETARCH}-cross && \
     ln -s /usr/bin/clang-15 /usr/bin/clang
 
 WORKDIR /pwru
 COPY . .
-RUN make local-release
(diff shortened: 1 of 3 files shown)


akerouanton:
Hm it seems I can't compile the amd64 image from my macbook, although it works properly on a Linux machine with a stock Docker Engine. I'll try to fix that.

jschwinger233 (approved):
Thank you!

@brb You know this better than me, how do you like it?

brb (approved):
Thanks!

brb:
@akerouanton I was informed by @aanm that Docker Hub, where we publish `cilium/pwru` images, does not support the multi-arch.

akerouanton:
> Docker Hub [...] does not support the multi-arch.

What do you mean @aanm? Docker Hub itself _does_ support multi-arch images. For instance: https://hub.docker.com/r/albinkerouanton006/pwru/tags.

aanm:
@brb @akerouanton what I meant was that hub.docker.com does not **build** multi-arch images but it's fine to publish.

akerouanton:
@aanm I see. I think a [build hook](https://docs.docker.com/docker-hub/builds/advanced/#build-hook-examples) could do the job, but Cilium needs to have either a Docker Pro, Team, or Business subscription as is noted at the top of that doc page. Is that the case?

aanm:
> @aanm I see. I think a [build hook](https://docs.docker.com/docker-hub/builds/advanced/#build-hook-examples) could do the job, but Cilium needs to have either a Docker Pro, Team, or Business subscription as is noted at the top of that doc page. Is that the case?

@akerouanton no, that's not the case. All multi-arch builds that Cilium has are done via GitHub actions. For example [here](https://github.com/cilium/cilium/blob/ca18209dd67a5b6e8e3cf6b114441cc9495d37e3/.github/workflows/build-images-ci.yaml#L35)

Issue #448, opened 2024-10-29
Improve observability of bpf map operations


jschwinger233:

jschwinger233:
I'm going to work out a simple version: with a flag, pwru prints map->name with key and value in hex.

PR #453, opened 2024-11-08
[Draft] Collect bpf helper arguments related to bpf map

jschwinger233:
The idea is to identify bpf map-related helpers in userspace, then use kprobe cookie to notify bpf program whether this requires arguments collecting.

There are at least following steps:
1. collect map->name. We assume `bpf_map *` is always at PT_REGS_PARM1(ctx).
2. collect key buffer. Key size can be determined by map->key_size, then we can simply call bpf_probe_read_kernel to collect the buffer. BTW we assume `void *key` is always at PT_REGS_PRAM2(ctx).
3. collect value buffer. This is a bit harder because it can be at retval for bpf_map_lookup_elem(). So userspace pwru must tell this info to bpf via cookie:
   a. if value is in the return (bpf_map_lookup): bpf_probe_read_kernel(ctx->rax) at kretprobe to collect it
   b. if value is in the param (bpf_map_update): bpf_probe_read_kernel(PT_REGS_PARM3(ctx)) to collect it
4. Similar to stack backtrace and shinfo, we'll add a new member `print_bpf_map_id` to event.

Some unexpected work may happen, such as bpf stack being used up. If that tragedy takes place, I'll use a per-cpu map to save bpf stack. 

This doesn't require BTF at all, but if necessary we can extend it to support BTF parsing.

Fixes: #448


jschwinger233:
I hope this helps me understand how Cilium CT works. Even after 18 months since onboarding, CT is still a mystery to me.

brb:
:+1: the idea

jschwinger233:
I labelled this "don't merge' because it's on the top of https://github.com/cilium/pwru/pull/477 which is pending.

brb:
#477 has been merged. Could your rebase? Thanks

bpf/kprobe_pwru.c
```diff
@@ -340,7 +368,7 @@ static __always_inline u64
 sync_fetch_and_add(void *id_map) {
 	u32 *id = bpf_map_lookup_elem(id_map, &ZERO);
 	if (id)
-		return ((*id)++) | ((u64)bpf_get_smp_processor_id() << 32);
+		return ((*id)++) | ((u64)(bpf_get_smp_processor_id() + 1) << 32);
```
jspaleta, line 371: Is this an off by one bug in the existing codebase?

PR #455, opened 2024-11-10
bpf: Remove OUTPUT_SKB&HAS_KPROBE_MULTI macros

Asphaltt:
When call set_skb_btf() or set_shinfo_btf(), these functions are protected by constant CFG.

Then, while loading kprobe_pwru into kernel, if no --output-skb/--output-skb-shared-info, set_skb_btf()/set_shinfo_btf() won't be verified. Next, these unreachable code will be eliminated by verifier.

### DIFF
--- a/internal/pwru/btf.go
+++ b/internal/pwru/btf.go
@@ -1,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0
+/* Copyright Leon Hwang */
+/* Copyright Authors of Cilium */
+
+package pwru
+
+import (
+	"fmt"
+
+	"github.com/cilium/ebpf/btf"
+)
+
+func GetStructBtfID(btfSpec *btf.Spec, structName string) (btf.TypeID, error) {
+	types, err := btfSpec.AnyTypesByName(structName)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get BTF types by name %s: %w", structName, err)
+	}
+
+	for _, t := range types {
+		if s, ok := t.(*btf.Struct); ok {
+			return btfSpec.TypeID(s)
+		}
+	}
+
+	return 0, fmt.Errorf("struct %s not found", structName)
+}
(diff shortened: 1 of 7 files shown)


jschwinger233:
Just for the record: 
1. 6.1 and bpf-next see `pwru --backend kprobe-multi` broken while `--backend kprobe` stil works.
2. 5.4 fails due to CO-RE issues on sk_buff. My guess is some members are deleted / renamed between 5.4 and 5.15.

internal/pwru/btf.go
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: Apache-2.0
+/* Copyright Leon Hwang */
```
jschwinger233, line 2: Do you mind adding a line of `/* Copyright Authors of Cilium */` below, like https://github.com/cilium/pwru/blob/main/internal/pwru/types.go#L3 ?

bpf/features.c
```diff
@@ -0,0 +1,26 @@
...
+{
+	features.kprobe_happened = true;
+
+	/* Detect if bpf_snprintf_btf is available.
+	 * Since: c4d0bfb45068 ("bpf: Add bpf_snprintf_btf helper")
+	 */
+	features.has_snprintf_btf = bpf_core_enum_value_exists(enum bpf_func_id, BPF_FUNC_snprintf_btf);
+
+	return BPF_OK;
+}
```
jschwinger233, line 24: Can we use a similar approach to https://github.com/cilium/pwru/blob/v1.0.9/internal/pwru/utils.go#L138 for feature detection?

bpf/features.c
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/* Copyright Leon Hwang */
+
```
jschwinger233, line 3: Same Copyright issue.

jschwinger233 (commented):
Always love to see code simplification! Left some comments to discuss.

bpf/features.c
Asphaltt: I'll have a try.

brb (approved):
Thanks!

Issue #295, opened 2023-12-08
`cb[]` (array of values) is not printed correctly when dumping the full skb


qmonnet:
When running pwru with `--output-skb`, we dump the full content of the socket buffer. However, the value for `cb` (`char cb[48]`) seems incorrect. I saw only the first value from the array (corresponding the first `char`), the following ones were not displayed.

I think `cb` is the only array in the skb, but this would probably related to the way we handle arrays in general.

[context](https://github.com/cilium/cilium/issues/29351#issuecomment-1846476500)

jschwinger233:
Considering that Cilium heavily relies on skb->cb, I can't tolerate this issue any longer! 

Do you think it's a good idea to fetch and output skb->cb when `--output-meta` is set? Or we add a new flag `--output-skb-cb`? Any suggestion?

The best solution looks like https://github.com/cilium/pwru/issues/11, but it requires huge engineering...

qmonnet:
Not sure what #11 mean, do you mean finding a way to specify what fields from the skb we want to print? An alternative to that would be to output the SKB metadata (possibly the other info, too) in JSON and filter with `jq`?

```
Do you think it's a good idea to fetch and output `skb->cb` when --output-meta is set? Or we add a new flag `--output-skb-cb`? Any suggestion?
```

I personally wouldn't mind printing the `skb->cb` by default. We'd need to fix `bpf_snprintf_btf()` to retrieve the content correctly though, if I understand correctly?

jschwinger233:
> We'd need to fix bpf_snprintf_btf() to retrieve the content correctly though, if I understand correctly?

True, although it would be long way to go.

brb:
Maybe as a temp solution we could add `skb->cb[]` to `--output-meta`'s output if `--filter-trace-tc` is set? The latter is expected for tracing BPF progs.

PR #461, opened 2024-12-04
Collect and output skb->cb when --filter-trace-tc

jschwinger233:
According to kernel verifier implementation[1], __sk_buff->cb[0:5] will be mapped to sk_buff->cb[8:28]. We'll collect sk_buff->cb[8:28] in bpf then cast u8[20] to u32[5] in userspace for output.

[1] https://elixir.bootlin.com/linux/v6.8/source/net/core/filter.c#L9593

Fixes: https://github.com/cilium/pwru/issues/295


bpf/kprobe_pwru.c
```diff
@@ -443,6 +444,10 @@ set_output(void *ctx, struct sk_buff *skb, struct event_t *event) {
 	if (cfg->output_stack) {
 		event->print_stack_id = bpf_get_stackid(ctx, &print_stack_map, BPF_F_FAST_STACK_CMP);
 	}
+
+	if (cfg->output_cb) {
+		bpf_probe_read_kernel(&event->meta.cb, sizeof(event->meta.cb), (void *)&skb->cb[8]);
```
Asphaltt, line 449: It's hard to understand `skb->cb[8]` here. It's better to convert `skb->cb` as `struct qdisc_skb_cb *`, and then read data from `cb->data` like:
```C
struct qdisc_skb_cb *cb = (typeof(cb))(void *) skb->cb;
bpf_probe_read_kernel(&event->meta.cb, sizeof(event->meta.cb), (void *) &cb->data);
```

bpf/kprobe_pwru.c
```diff
@@ -54,7 +54,7 @@ struct skb_meta {
 	u32 len;
 	u32 mtu;
 	u16 protocol;
-	u16 pad;
+	u8 cb[20];
```
Asphaltt, line 57: It's OK to define as `u32 cb[5]` like `cb` in `struct __sk_buff`, I think.

Asphaltt (changes requested):
LGTM,  I’ve left two comments for further discussion.

Asphaltt (approved):
LGTM.

brb (approved):
Good find, thanks!

Issue #394, opened 2024-07-07
Add flag --trace-only-bpf to trace only BPF (TC/XDP) progs


brb:

brb:
@sachintiptur Yep, go for it. Thanks.

PR #469, opened 2024-12-16
Add --filter-trace-only-bpf

sachintiptur:
Add --filter-trace-only-bpf flag to trace only TC and XDP programs. 
Fixes [#394 ]

### DIFF
--- a/main.go
+++ b/main.go
@@ -139,6 +139,13 @@ func main() {
 		}
 	}
 
+	// Skip the skb funcs when --filter-trace-only-bpf is enabled
+	if flags.FilterTraceOnlyBpf {
+		for i := 1; i <= 5; i++ {
+			delete(bpfSpec.Programs, fmt.Sprintf("kprobe_skb_%d", i))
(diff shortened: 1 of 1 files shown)


jschwinger233 (commented):
I don't think this achieved the goal. Running `pwru --filter-trace-only-bpf 'dst host 1.1.1.1'` built from this PR also outputs tracing results from normal skb functions. Unless I misunderstand the original issue https://github.com/cilium/pwru/issues/394, the expected behavior is to attach bpf programs **only**, rather than normal skb funcitons.

jschwinger233:
@sachintiptur By default, pwru traces events from skb functions, such as `ip_rcv`. When setting `--filter-trace-xdp`, pwru **also** traces events from xdp bpf prog; when settings `--filter-trace-tc`, pwru **also** traces events from tc bpf prog.

Your PR that enables both flags makes pwru trace events from 1)skb functions + 2)xdp bpf + 3)tc bpf. The `--filter-trace-only-bpf` is supposed to trace 2)xdp bpf + 3)tc bpf **only**.

Issue #462, opened 2024-12-04
pwru fails to parse helper symbols when --filter-track-bpf-helpers


jschwinger233:
```
$ s pwru --filter-mark 0x78540a00  --filter-track-bpf-helpers        'dst host 10.244.1.136 and dst port 8080 and tcp[tcpflags]=tcp-syn' 
2024/12/05 02:12:02 Attaching kprobes (via kprobe-multi)...
1641 / 1641 [-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00% ? p/s
2024/12/05 02:12:02 Attached (ignored 0)
2024/12/05 02:12:02 Listening for events..
SKB                CPU PROCESS          NETNS      MARK/x        IFACE       PROTO  MTU   LEN   TUPLE FUNC
0xffff9296e13e3ce8 5   ~um-envoy:491632 4026533895 78540a00        0         0x0000 1500  60    10.244.1.95:42104->10.244.1.136:8080(tcp) ip_local_out
0xffff9296e13e3ce8 5   ~um-envoy:491632 4026533895 78540a00        0         0x0000 1500  60    10.244.1.95:42104->10.244.1.136:8080(tcp) __ip_local_out
0xffff9296e13e3ce8 5   ~um-envoy:491632 4026533895 78540a00        0         0x0800 1500  60    10.244.1.95:42104->10.244.1.136:8080(tcp) nf_hook_slow
0xffff9296e13e3ce8 5   ~um-envoy:491632 4026533895 78540a00        0         0x0800 1500  60    10.244.1.95:42104->10.244.1.136:8080(tcp) ip_output
0xffff9296e13e3ce8 5   ~um-envoy:491632 4026533895 78540a00  cilium_host:3   0x0800 1500  60    10.244.1.95:42104->10.244.1.136:8080(tcp) nf_hook_slow
0xffff9296e13e3ce8 5   ~um-envoy:491632 4026533895 78540a00  cilium_host:3   0x0800 1500  60    10.244.1.95:42104->10.244.1.136:8080(tcp) apparmor_ip_postroute
0xffff9296e13e3ce8 5   ~um-envoy:491632 4026533895 78540a00  cilium_host:3   0x0800 1500  60    10.244.1.95:42104->10.244.1.136:8080(tcp) ip_finish_output
[...]

PR #477, opened 2025-01-03
Events from kprobe.multi have eventTypeKprobeMulti set in .type

jschwinger233:
Even with --backend kprobe-multi, pwru still uses kprobe for --filter-track-bpf-helpers. This patch makes pwru capable of distinguishing event types in order to adjust addressres for symbol resolution.

Fixes: https://github.com/cilium/pwru/issues/462


brb (approved):
Good find, thanks!

PR #484, opened 2025-01-15
Fix broken --filter-trace-xdp caused by go-ebpf v0.17.1

### LINKED ISSUE #483: Failed to trace XDP: using replacement map .bss: Flags: 1024 changed to 0: map spec is incompatible with existing map
```bash
# ./pwru --backend kprobe --filter-trace-tc --filter-trace-xdp --filter-func '.*icmp.*' icmp
2025/01/15 14:04:34 Attaching tc-bpf progs...
2025/01/15 14:04:34 Attaching xdp progs...
2025/01/15 14:04:34 failed to trace XDP progs: failed to trace bpf progs: failed to load objects:
using replacement map .bss: Flags: 1024 changed to 0: map spec is incompatible with existing map
```

I add a log to check `.bss` flags:

```
2025/01/15 14:36:35 spec .bss flags: 0x0, .bss map flags: 0x400
```

`0x400` is **BPF_F_MMAPABLE**.
```c
/* Enable memory-mapping BPF map */
	BPF_F_MMAPABLE		= (1U << 10),
```

It seems this flag is set by go-ebpf v0.17.1.

I'll fix it asap.

Asphaltt:
Fixes #483 

There're two issue:

One is `BPF_PROG_ADDR` must be set via `spec.Variable`.

Another one is `.bss` map flag must be updated because go-ebpf v0.17.1 set `BPF_F_MMAPABLE` flag to created `.bss` map. As a result, we can reuse `.bss` map.

Here's the test result:

```bash
2025/01/15 15:14:26 Attaching tc-bpf progs...
2025/01/15 15:14:26 Attaching xdp progs...
2025/01/15 15:14:26 Attaching kprobes (via kprobe)...
29 / 29 [--------------------------------------------------------------------------------------------------------------------------------------] 100.00% ? p/s
2025/01/15 15:14:26 Attached (ignored 0)
2025/01/15 15:14:26 Listening for events..
SKB                CPU PROCESS          NETNS      MARK/x        IFACE       PROTO  MTU   LEN   __sk_buff->cb[]                                          TUPLE FUNC
0xffffb5dd40600b00 6   <empty>:0        4026531840 0            ens33:2      0x0000 1500  98    [0x00000000,0x00000000,0x00000000,0x00000000,0x00000000] 192.168.241.1:0->192.168.241.133:0(icmp) bpf_prog_3b185187f1855c4c_dummy[bpf](xdp)
0xffff995b7756d400 6   <empty>:0        4026531840 0            ens33:2      0x0800 1500  98    [0x00000000,0x00000000,0x00000000,0x00000000,0x00000000] 192.168.241.1:0->192.168.241.133:0(icmp) bpf_prog_e20a685998fd41c8_entry1[bpf](tc)
0xffff995b7756d400 6   <empty>:0        4026531840 0            ens33:2      0x0800 65536 64    [0x00000000,0x00000000,0x00000000,0x00000000,0x00000000] 192.168.241.1:0->192.168.241.133:0(icmp) icmp_rcv
0xffff995b7756d400 6   <empty>:0        4026531840 0            ens33:2      0x0800 65536 56    [0x00000000,0x00000000,0x00000000,0x00000000,0x00000000] 192.168.241.1:0->192.168.241.133:0(icmp) icmp_echo
0xffff995b7756d400 6   <empty>:0        4026531840 0            ens33:2      0x0800 65536 56    [0x00000000,0x00000000,0x00000000,0x00000000,0x00000000] 192.168.241.1:0->192.168.241.133:0(icmp) icmp_reply
^C2025/01/15 15:14:30 Received signal, exiting program..
2025/01/15 15:14:30 Detaching kprobes...
29 / 29 [------------------------------------------------------------------------------------------------------------------------------------] 100.00% 132 p/s
```

And there's only one `.bss` map:

```bash
# bpftool m | grep -c .bss
1
```

### DIFF
--- a/internal/pwru/tracing.go
+++ b/internal/pwru/tracing.go
@@ -92,10 +92,8 @@ func (t *tracing) traceProg(spec *ebpf.CollectionSpec,
 	}
 
 	spec = spec.Copy()
-	if err := spec.RewriteConstants(map[string]any{
-		"BPF_PROG_ADDR": addr,
-	}); err != nil {
-		return fmt.Errorf("failed to rewrite bpf prog addr: %w", err)
+	if err := spec.Variables["BPF_PROG_ADDR"].Set(addr); err != nil {
+		return fmt.Errorf("failed to set bpf prog addr: %w", err)
 	}
 
 	spec.Programs[tracingName].AttachTarget = prog
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -30,7 +30,7 @@
 const static bool TRUE = true;
 const static u32 ZERO = 0;
 
-volatile const static __u64 BPF_PROG_ADDR = 0;
+volatile const __u64 BPF_PROG_ADDR = 0;
 
 enum {
 	TRACKED_BY_FILTER = (1 << 0),


internal/pwru/tracing.go
```diff
@@ -136,6 +134,10 @@ func (t *tracing) trace(coll *ebpf.Collection, spec *ebpf.CollectionSpec,
 	delete(replacedMaps, ".rodata")
 	opts.MapReplacements = replacedMaps
 
+	if m, ok := coll.Maps[".bss"]; ok && spec.Maps[".bss"].Flags != m.Flags() {
```
ti-mo, line 137: Just curious, why modify .bss directly?

Asphaltt: Before the introduction of `mmap` support for `.bss` maps, the flags were not set as **BPF_F_MMAPABLE** and remained **0**, as specified in the map's specification.

When using `mmap` with a `.bss` map, the flags must be set to **BPF_F_MMAPABLE** to indicate to the kernel that the map utilizes `mmap` memory.

To enable the reuse of `.bss` maps, it is necessary to update the flags in the map specification.

Additionally, I plan to introduce a global `ready` variable in BPF to control whether packet tracing is enabled. This approach aims to mitigate the performance overhead of starting and stopping pwru with over 1,000 kprobes, which can be significantly slow.

ti-mo: Okay, but why reuse .bss (are you loading multiple Collections?)?

You're probably referring to this, but note that the lib sets mmapable for you if the kernel supports it: https://github.com/cilium/ebpf/blob/f224bff41d4d89fca11ddac24453f6dacffb6f89/collection.go#L511. If not, `Collection.Variables` will be empty.

Asphaltt: > Okay, but why reuse .bss (are you loading multiple Collections?)?

Yes, the new Collection is used for tracing XDP or tc-BPF programs and reuses the `.bss` map.

> You're probably referring to this, but note that the lib sets mmapable for you if the kernel supports it

Correct.

I verified that **BPF_F_MMAPABLE** has been supported since commit torvalds/linux@fc9702273e2e ("bpf: Add mmap() support for BPF_MAP_TYPE_ARRAY") introduced in the v5.5 kernel.

Similarly, tracing was introduced in commit torvalds/linux@fec56f5890d9 ("bpf: Introduce BPF trampoline"), also in the v5.5 kernel. Therefore, **BPF_F_MMAPABLE** and tracing are expected to be supported concurrently starting with v5.5.

On a related note, the `ready` global variable can use a standalone map named `.data.ready` if compatibility with v5.4 kernels is required.

ti-mo: > I verified that BPF_F_MMAPABLE has been supported since commit https://github.com/torvalds/linux/commit/fc9702273e2e ("bpf: Add mmap() support for BPF_MAP_TYPE_ARRAY") introduced in the v5.5 kernel.

This still doesn't require you to set the flag manually, the lib will do it for you. I'd recommend you don't touch this.

> On a related note, the ready global variable can use a standalone map named .data.ready if compatibility with v5.4 kernels is required.

I'd recommend you emit a ready variable to a section you control (`volatile bool ready __attribute__((section(".data.ready")))`), and only share the `.data.ready` map between Collections via `opts.MapReplacements`. You can still use `Collection{,Spec}.Variables["ready"]` to modify the variable after loading, it'll point to the same underlying map.

 `.bss` is where all zero-initialized variables go, including local variables, so I'm not sure why you assume this is safe to share between Collections originating from different objects (I assume you have separate tc/xdp/kprobe objects). I would also steer clear of sharing .rodata maps between Collections, that's inviting subtle bugs. Just apply the same map of values to each CollectionSpec.Variables.

Asphaltt: > This still doesn't require you to set the flag manually, the lib will do it for you. I'd recommend you don't touch this.

I don't want to touch it. But it will fail without updating the flags.

```bash
# ./pwru --filter-trace-tc --filter-trace-xdp --filter-func '.*icmp.*' icmp
2025/01/23 12:27:49 Attaching tc-bpf progs...
2025/01/23 12:27:49 failed to trace TC progs: failed to trace bpf progs: failed to load objects:
using replacement map .bss: Flags: 1024 changed to 0: map spec is incompatible with existing map
```

Is the lib responsible for this issue? WDYT?

> I'd recommend you emit a ready variable to a section you control (`volatile bool ready __attribute__((section(".data.ready"))))`, and only share the `.data.ready` map between Collections via `opts.MapReplacements`. You can still use `Collection{,Spec}.Variables["ready"]` to modify the variable after loading, it'll point to the same underlying map.

Sure. Thank you for your recommendation.

> I'm not sure why you assume this is safe to share between Collections originating from different objects (I assume you have separate tc/xdp/kprobe objects).

Yes, there are separate tc/xdp/kprobe collections in pwru.

Reusing `.bss` map is to ensure that there is only one pwru configs between these collections. And the `.bss` map populates data in `main.go` before tracing tc/xdp progs.

---

But when I confirm global variables in `.bss` map, I think these global variables are used incorrectly. See #491.

ti-mo: > Is the lib responsible for this issue? WDYT?

Interesting! We're setting the mmapable flag on the spec after checking for map compatibility, which is obviously not going to work. I'll move this around and ship it in a 0.17.2, we've squashed some other bugs in the mean time as well. I'll report back here.

Asphaltt: > We're setting the mmapable flag on the spec after checking for map compatibility, which is obviously not going to work.

In pwru, the spec is copied before creating collection. So, if the mmapable flag is set while creating collection, it won't work for pwru. It would be better to check map compatibility without mmapable flag for `.bss`/`.data` maps.

ti-mo: Just submitted https://github.com/cilium/ebpf/pull/1655, please give it a try!

brb: @ti-mo Thanks for fixing. When do you expect to release v0.17.2?

ti-mo: Should be out today/tomorrow!

brb: @Asphaltt https://github.com/cilium/ebpf/releases/tag/v0.17.2 is out! Could you update your PR?
```
go get github.com/cilium/ebpf@v0.17.2
go mod vendor
go mod tidy
```

Asphaltt: I'll update it after merging the PR #495

brb: Thank you! Done.

brb (approved):
Thank you!

Issue #488, opened 2025-01-21
missing constants from .rodata when --filter-trace-tc


smagnani96:
Upon running `pwru --filter-trace-tc`, I receive the following error:

```bash
Attaching tc-bpf progs...
failed to trace TC progs: failed to trace bpf progs: failed to rewrite bpf prog addr: rewrite constants: some constants are missing from .rodata: BPF_PROG_ADDR
```

This might be due to the changes to the new ebpf Variable API + some map replacements while calling `func (t *tracing) traceProg(spec *ebpf.CollectionSpec, 	opts *ebpf.CollectionOptions, prog *ebpf.Program, n2a BpfProgName2Addr, tracingName string, ) error`

jschwinger233:
It's https://github.com/cilium/pwru/pull/481 😬

jschwinger233:
(need more test in ci 〰

smagnani96:
This is a duplicate of #483 (error in subsequent steps to solve this issue is the same).


> (need more test in ci 〰

Should open a new Issue?

PR #489, opened 2025-01-21
fix missing constant from .rodata in --filter-trace-tc

smagnani96:
While running pwru with `--filter-trace-tc`, I received the following error:

```bash
Attaching tc-bpf progs...
failed to trace TC progs: failed to trace bpf progs: failed to rewrite bpf prog addr: rewrite constants: some constants are missing from .rodata: BPF_PROG_ADDR
```

I then decided to drop the static identifier from BPF_PROG_ADDR and switch from RewriteConstants to the new global variable API. However, this was not enough, since I started receiving:

```bash
Attaching tc-bpf progs...
failed to trace TC progs: failed to trace bpf progs: failed to load objects:
using replacement map .bss: Flags: 1024 changed to 0: map spec is incompatible with existing map
```

Therefore, I've added a line to delete also `.bss` maps other than `.rodata` while rewriting specs. This enables back running pwru with `--filter-trace-tc`.

Fixes: #488

(This will be my 1st pwru PR, please any kind of suggestion/feedback is very welcome)

### DIFF
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -30,7 +30,7 @@
 const static bool TRUE = true;
 const static u32 ZERO = 0;
 
-volatile const static __u64 BPF_PROG_ADDR = 0;
+volatile const __u64 BPF_PROG_ADDR = 0;
 
 enum {
 	TRACKED_BY_FILTER = (1 << 0),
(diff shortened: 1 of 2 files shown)


jschwinger233 (approved):

Asphaltt:
Dup PR #484 

It's urgent to fix the issue and add ci-test for `--filter-trace-tc` and `--filter-trace-xdp`. cc @brb @jschwinger233

smagnani96:
> Dup PR #484

Thanks for pointing out, closing this :pray:

PR #492, opened 2025-01-23
Add dummy --filter-track-{tc,xdp} tests

brb:
Depends on https://github.com/cilium/pwru/pull/484


.github/workflows/test.yml
```diff
@@ -36,14 +36,25 @@ jobs:
...
+
   test:
     runs-on: ubuntu-latest
     name: Test
     needs: build
     strategy:
       fail-fast: false
       matrix:
-        kernel: [ '5.4-20241218.004849', '5.10-20241218.004849', '5.15-20241218.004849', '6.1-20241218.004849', '6.6-20241218.004849', '6.12-20241218.004849', 'bpf-next-20250105.013256' ]
+         kernel: [ '5.4-20241218.004849', '5.10-20241218.004849', '5.15-20241218.004849', '6.1-20241218.004849', '6.6-20241218.004849', '6.12-20241218.004849', 'bpf-next-20250105.013256' ]
```
tklauser, line 57: Unintended whitespace change?

.github/workflows/test.yml
```diff
@@ -36,6 +36,17 @@ jobs:
         name: pwru
         path: pwru
 
+    - name: Build test-app
+      run: |
+        cd test-app
+        go build .
```
tklauser, line 42: I wonder whether we should run `go generate` here to generate `bpf_x86_bpfel.*` rather than adding these generated files to the repo. But then again I'm not sure what the standard practice is regarding these generated files for ebpf-go programs.

brb: Yeah, I was considering that, but then from the commit msg:

>  This commit adds the compiled BPF progs, to reduce the runtime dependency burden for the GH workflows (no need to install Clang)

I am expecting that prog to change very rarely, so the no need to install Clang for each PR outweighs the convenience.

tklauser: Should have read the commit message 😬 all good then

tklauser (approved):

PR #509, opened 2025-03-19
[RFC] pwru: Support --output-skb-metadata

### LINKED ISSUE #502: Support dumping skb hash
Extend pwru to support duming:

- skb hash
- whether the skb hash is L4 hash (vs L3 or lower)
- whether the skb hash is SW hash (vs HW)

It would be useful to see this meta data at each traced function.

For example, packet comes in from HW/NIC, gets into Cilium, is rewritten (K8s services) and pushed via tunnel device to a backend. It would be useful to analyze the skb hash at each of these stages.

Asphaltt:
Fixes #502 

I'm porting such feature from [btrace](https://github.com/leonhwangprojects/btrace)'s `--output-arg 'skb->hash'`.

---

It's really useful to output specified metadatas of skb instead of outputing the whole skb btf data.

However, it's not easy to `--output-skb-metadata 'skb->mark'` because the metadata is provided by users.

Here're the steps to achieve it:

1. Parse the text to C AST.
2. Convert the C AST to bpf insns with `skb` btf type info.
3. Inject the `set_skb_metadata` stub with the generated insns.
4. Output the metadata with the provided field's btf type info.

e.g.

```bash
$ sudo ./pwru --filter-func 'icmp_rcv' --output-skb-metadata 'skb->hash' --output-skb-metadata 'skb->sw_hash' --output-skb-metadata 'skb->l4_hash' --output-skb-metadata 'skb->mark' host 1.1.1.1
2025/03/19 16:27:18 Attaching kprobes (via kprobe-multi)...
1 / 1 [----------------------------------------------------------------------------------------------------------------------------------------] 100.00% ? p/s
2025/03/19 16:27:18 Attached (ignored 0)
2025/03/19 16:27:18 Listening for events..
SKB                CPU PROCESS          NETNS      MARK/x        IFACE       PROTO  MTU   LEN   TUPLE FUNC
0xffff991f3ec49600 6   <empty>:0        4026531840 0            ens33:2      0x0800 65536 64    1.1.1.1:0->192.168.241.133:0(icmp) icmp_rcv hash=0 sw_hash=0 l4_hash=0 mark=0
^C2025/03/19 16:32:56 Received signal, exiting program..
2025/03/19 16:32:56 Detaching kprobes...
1 / 1 [----------------------------------------------------------------------------------------------------------------------------------------] 100.00% ? p/s
```


brb:
Yep, this feature makes sense :+1: 

Mind updating the PR?

internal/pwru/skb_metadata.go
```diff
@@ -0,0 +1,163 @@
...
+func isValidChar(c byte) bool {
+	return isChar(c) || c == '_' || isDigit(c)
+}
+
+func newSkbMetadata(expr string, spec *btf.Spec) (*SkbMetadata, error) {
+	var md SkbMetadata
+	md.expr = expr
+
+	if !strings.HasPrefix(expr, "skb") {
+		return nil, errors.New("--output-skb-metadata must starts with skb")
```
brb, line 51: Nit: s/starts/start/

internal/pwru/skb_metadata.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	for i := len(expr) - 1; i >= 0; i-- {
+		if !isValidChar(expr[i]) {
```
brb, line 55: Why can't we return an error instead?

internal/pwru/skb_metadata.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func InjectSetSkbMetadata(prog *ebpf.ProgramSpec, md []*SkbMetadata) error {
+	if len(md) == 0 {
+		return nil
+	}
+
+	var insns asm.Instructions
+	if len(md) > 1 {
+		insns = append(insns,
```
brb, line 123: Why is it not needed when len(md) = 1?

internal/pwru/xdp_metadata.go
```diff
@@ -0,0 +1,148 @@
+// SPDX-License-Identifier: Apache-2.0
```
brb, line 1: Maybe we could make skb_metadata.go generic, so that it can be reused by XDP?

brb (commented):
Nice! Mostly questions / nits.

internal/pwru/skb_metadata.go
Asphaltt: Ack.

internal/pwru/skb_metadata.go
Asphaltt: This `for` is to get the last attribute name, e.g. 'ifindex' for 'skb->dev->ifindex'.
Afterwards, while outputting its value, the output should be 'ifindex=XX' instead of 'skb->dev->ifindex=XX'.

Besides it, I should add a `for` to validate all chars.

internal/pwru/skb_metadata.go
Asphaltt: if 'len(md) == 1', the **R1** already is `skb`.

As for 'md[1:]', it has to backup and recover **R1** with other register like **R7**.

internal/pwru/xdp_metadata.go
Asphaltt: Sure. I will do it.

brb (approved):
Thanks!

jschwinger233 (approved):

PR #543, opened 2025-04-15
Use jschwinger233/elibpcap to compile and inject libpcap

jschwinger233:
The https://github.com/jschwinger233/elibpcap has been adopted by several popular projects such as https://github.com/gojue/ecapture, let's use it to simplify the codebase.

Historically, elibpcap was derived from pwru/internal/libpcap/ following pwru#198. The two share nearly identical code and logic, so there should be no compatibility issues in transitioning pwru to use elibpcap.

Signed-off-by: gray <gray.liang@isovalent.com>


Makefile
```diff
@@ -12,14 +12,16 @@ LIBPCAP_ARCH ?= x86_64-unknown-linux-gnu
 # For compiling libpcap and CGO
 CC ?= gcc
 ARCHS ?= amd64 arm64
+CGO_CFLAGS="-I/pwru/libpcap"
+CGO_LDFLAGS="-L/pwru/libpcap -lpcap -static"
```
chenhengqi, line 16: Failed to compile locally, the path should be `./libpcap`.

jschwinger233: What command did you run locally? `make release` and `make` are both okay on my local ubuntu.

chenhengqi: I don't know why this is supposed to work locally. The CI build with docker which set workdir to /pwru so there is no problem.

rgo3: I agree that this shouldn't be an absolute path, otherwise e.g. `make pwru` will fail on hosts where `/pwru/libpcap/` does not exist. We could also consider using `-L$(CURDIR)/libpcap`.

brb:
If I run `make`, I get the following error:
```
CGO_CFLAGS="-I/pwru/libpcap" CGO_LDFLAGS="-L/pwru/libpcap -lpcap -static" CC=cc GOARCH=amd64 CGO_ENABLED=1 go build  \
        -ldflags "-w -s \
        -X 'github.com/cilium/pwru/internal/pwru.Version=v1.0.10-pre-69-g3c948a7'"
# github.com/jschwinger233/elibpcap
vendor/github.com/jschwinger233/elibpcap/compile.go:15:10: fatal error: pcap.h: No such file or directory
   15 | #include <pcap.h>
      |          ^~~~~~~~
compilation terminated.
make: *** [Makefile:24: pwru] Error 1
```

The issue goes away if I apply the following diff:
```
diff --git a/vendor/github.com/jschwinger233/elibpcap/compile_static.go b/vendor/github.com/jschwinger233/elibpcap/compile_static.go
index 928555f..366d101 100644
--- a/vendor/github.com/jschwinger233/elibpcap/compile_static.go
+++ b/vendor/github.com/jschwinger233/elibpcap/compile_static.go
@@ -1,12 +1,12 @@
+//go:build static || !dynamic
 // +build static !dynamic

 package elibpcap

-
 /*
-#cgo LDFLAGS: -L/usr/local/lib -lpcap -static
+#cgo CFLAGS: -I${SRCDIR}/../../../../libpcap
+#cgo LDFLAGS: -L${SRCDIR}/../../../../libpcap -lpcap -static
 #include <stdlib.h>
 #include <pcap.h>
 */
 import "C"
```

@jschwinger233 One solution would be to ship the libcap in the [jschwinger233/elibpcap](https://github.com/jschwinger233/elibpcap)?

brb:
> One solution would be to ship the libcap in the [jschwinger233/elibpcap](https://github.com/jschwinger233/elibpcap)?

@jschwinger233 :wave: Would you accept such a contribution?

PR #552, opened 2025-05-15
Add percpu buffer flag

### LINKED ISSUE #508: --output-skb is cut short
Example: [pwru.log](https://github.com/user-attachments/files/19128055/pwru.log)

Started as `pwru --output-skb --filter-trace-tc 'port 80'`.

Version: [Release v1.0.10-pre](https://github.com/cilium/pwru/releases/tag/untagged-b7ad3f3c997fe992fe3e)

AdvH039:
Tries to fix #508
(Edit : What command to use for linting?)

### DIFF
--- a/internal/pwru/types.go
+++ b/internal/pwru/types.go
@@ -65,6 +65,8 @@ type Flags struct {
 	ReadyFile string
 
 	Backend string
+
+	SetPerCPUBuf uint32
 }
 
 func (f *Flags) SetFlags() {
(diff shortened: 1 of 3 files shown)


brb:
Thanks for the PR.

Unfortunately, we cannot use variable length arrays. One solution would be to use a BPF map with value `char foo[0]` for the buffer, and then set the value size with [bpf_map__set_value_size](https://libbpf.readthedocs.io/en/latest/api.html) from the agent before loading.

Please let me know if you are comfortable to implement the solution above.

AdvH039:
What do you mean by agent exactly?

brb:
The Go userspace process (main.go)

brb:
@ti-mo suggested:

> I'd prob use a simple percpu array and adjust `CollectionSpec.Maps["scratch"].ValueSize` as appropriate before loading

internal/pwru/types.go
```diff
@@ -110,6 +112,7 @@ func (f *Flags) SetFlags() {
 	flag.StringVar(&f.Backend, "backend", "",
 		fmt.Sprintf("Tracing backend('%s', '%s'). Will auto-detect if not specified.", BackendKprobe, BackendKprobeMulti))
 
+	flag.Uint32Var(&f.SetPerCPUBuf, "set-percpu-buf", 4096, "set the size of the percpu buffer")
```
brb, line 115: Maybe you could be more specific with the flag description? I.e. instead of saying "percpu buffer", you could say "set the size of buffers to print skb data (used by --output-skb and --output-skb-shared-info)".

brb (changes requested):
Thanks! One nit.

Issue #549, opened 2025-04-30
ebpf filter issue with `ip6 protochain 58`


mss:
I am trying to use this tool to debug some NDP issue. When I use the command `pwru ip6 protochain 58` I get this error:

```
2025/04/30 09:51:53 Failed to inject filter ebpf for kprobe_skb_4: unable to compute blocks: instruction 18: ja 4294967281 flows past last instruction
```

Just calling `pwru ip6 proto 58` works as intended (side issue: `icmp6` should be an alias for that one but results in a syntax error).

This is a Proxmox server running 6.8.12-9-pve.

brb:
Thanks for the issue.

Hit the same on the `6.11.0-25-generic` kernel. Looks like using `protochain` hits the BPF instruction limit. 

Is it possible to limit how "deep" the `protochain` should parse?

mss:
I don't think that's possible since theoretically it could be arbitrarily deep. The manpage says:
> The BPF code emitted by this primitive is complex and cannot be optimized by the BPF optimizer code, and is not supported by filter engines in the kernel, so this can be somewhat slow, and may cause more packets to be dropped.

Maybe the best solution (for now and if possible) would be to just disallow `protochain` with some kind of useful error?

brb:
Could you add the limitation to https://github.com/cilium/pwru/blob/main/KNOWN_ISSUES.md?

brb:
@jschwinger233 Good find (as always 😅)!  @mss Are you OK to document in the KNOWN_LIMITATIONS.md?

mss:
@brb yes, totally. Some documentation or improved error handling was why I raised this issue so the next person running into this will hopefully have a less puzzling day.

Just a suggestion: Maybe the command line parser or whatever feeds the arguments to the bpf compiler part could also bail out if it encounters the string `filterchain` as an argument?

brb:
Thanks!

>  Maybe the command line parser or whatever feeds the arguments to the bpf compiler part could also bail out if it encounters the string protochain as an argument?

:+1:

jschwinger233:
What I said in https://github.com/cilium/pwru/issues/549#issuecomment-2894245841 is not accurate, there are two issues, one in cloudflare.cbpfc, the other in libpcap, I opened pr/issue for them:
- https://github.com/cloudflare/cbpfc/pull/36
- https://github.com/the-tcpdump-group/libpcap/issues/1521

jschwinger233:
[...]

PR #558, opened 2025-05-23
Compile libpcap with --disable-protochain

jschwinger233:
There is a known issue that cloudfare/cbpfc can't properly handle "jump back" cBPF: https://github.com/cloudflare/cbpfc/pull/36. Unfortunately, "protochain" pcap filter often generates such "jump back" instructions, leading to errors like:

    unable to compute blocks: instruction 18: ja 4294967281 flows past last instruction

This patch disables protochain when compiling libpcap, so users receive a clearer error:

    $ pwru ip6 protochain 58
    2025/05/23 22:42:53 Failed to inject filter ebpf for kprobe_multi_skb_5: failed to pcap_compile 'ip6 protochain 58': protochain not supported (l3=false)

Issue: https://github.com/cilium/pwru/issues/549


brb (approved):
Thank you!

PR #569, opened 2025-06-19
ci: add workflow to build and push release multi-arch Docker images

### LINKED ISSUE #262: Multi-arch container image

tklauser:
Rather than letting Dockerhub build the images (which only suppports building amd64 in the subscription the Cilium project is using), use a GitHub actions workflow to build and push Docker images for amd64 and arm64 on release.

Fixes #262

### DIFF
--- a/.github/workflows/images.yaml
+++ b/.github/workflows/images.yaml
@@ -1,0 +1,48 @@
+name: Image Release Build
+
+on:
+  pull_request: {}
+  push:
+    tags:
+      - 'v*'
+
+concurrency:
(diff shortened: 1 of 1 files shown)


tklauser:
Looks like we need to define repo secrets for DockerHub username and password: https://github.com/cilium/pwru/actions/runs/15759403771/job/44422209509?pr=569

brb (approved):

Issue #559, opened 2025-06-05
pwru loongarch support?


vincentmli:
It appears pwru support loongarch, but when I build pwru on loongarch64 platform, I got:

```
[root@fedora pwru]# TARGET_GOARCH=loong64 GOARCH=loong64 make pwru
TARGET_GOARCH=loong64 go generate
Generating for loong64
/home/loongarch/pwru/bpf/kprobe_pwru.c:777:2: error: Unsupported architecture
  777 | #error Unsupported architecture
      |  ^
/home/loongarch/pwru/bpf/kprobe_pwru.c:823:14: error: use of undeclared identifier 'ip_offset'; did you mean 'Opt_offset'?
  823 |         return ip - ip_offset;
      |                     ^~~~~~~~~
      |                     Opt_offset
./bpf/headers/vmlinux-loongarch.h:104374:2: note: 'Opt_offset' declared here
 104374 |         Opt_offset = 7,
        |         ^
2 errors generated.
Error: compile: exit status 1
exit status 1
build.go:5: running "go": exit status 1
make: *** [Makefile:21: pwru] Error 1

```

@chenhengqi maybe `bpf/kprobe_pwru.c` need loongarch addition for `ip_offset` ?

```
766 static __always_inline u64
767 get_func_ip(void) {
768         u64 fp_tramp, ip;
769 
770 #if defined(bpf_target_x86)
771         static const int ip_offset = 5/* sizeof callq insn */;
772         u64 fp;
773 #elif defined(bpf_target_arm64)
774         /* Ref: commit b2ad54e1533e ("bpf, arm64: Implement bpf_arch_text_poke() for arm64") */
775         static const int ip_offset = 12/* sizeof 3 insns */;
776 #else
777 #error Unsupported architecture
778 #endif
```

Asphaltt:
It was caused by PR #468, which introduced a bunch of function IP enhancements for x86_64 and arm64, excluding loongarch.

vincentmli:
> It was caused by PR [#468](https://github.com/cilium/pwru/pull/468), which introduced a bunch of function IP enhancements for x86_64 and arm64, excluding loongarch.

@Asphaltt I guess you have fix for loongarch then?  hengqi says he may look at it too.

Asphaltt:
> I guess you have fix for loongarch then? 

Sorry, I'm unable to fix it, because I do not have loongarch device to test it. I'm unfamiliar with bpf on loongarch meanwhile.

[...]

PR #580, opened 2025-07-12
fix: pwru build error on LoongArch

vincentmli:
commit c4a911e ("Replace BPF_PROG_ADDR with get_func_ip()") caused pwru build error on LoongArch [0] because Leon do not have LoongArch to verify the change for LoongArch,so left out LoongArch.

Since LoongArch does not support bpf trampoline, get_func_ip is not going to be called on LoongArch, add a fake ip_offset value for LoongArch in get_func_ip to avoid the build error. We could fix the ip_offset value when LoongArch starts to support bpf trampoline.

[0]: https://github.com/cilium/pwru/issues/559

Suggested-by: Hengqi Chen <hengqi.chen@gmail.com>

### DIFF
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -773,6 +773,9 @@ get_func_ip(void) {
 #elif defined(bpf_target_arm64)
 	/* Ref: commit b2ad54e1533e ("bpf, arm64: Implement bpf_arch_text_poke() for arm64") */
 	static const int ip_offset = 12/* sizeof 3 insns */;
+#elif defined(bpf_target_loongarch)
+	/* LoongArch does not support trampoline, get_func_ip does not apply to LoongArch */
+	static const int ip_offset = 0;
 #else
 #error Unsupported architecture
 #endif


bpf/kprobe_pwru.c
```diff
@@ -773,6 +773,9 @@ get_func_ip(void) {
 #elif defined(bpf_target_arm64)
 	/* Ref: commit b2ad54e1533e ("bpf, arm64: Implement bpf_arch_text_poke() for arm64") */
 	static const int ip_offset = 12/* sizeof 3 insns */;
+#elif defined(bpf_target_loongarch)
+	/* LoongArch does not support trampoline, get_func_ip does not apply to LoongArch */
+	static const int ip_offset = 0;
```
chenhengqi, line 778: I am OK with this for now. We can revise it later once trampoline feature landed.

vincentmli:
Yes I tested on LoongArch hardware, I assume Github action could require some kind of LoongArch qemu instance, not familiar with how to run LoongArch qemu instance.

PR #600, opened 2025-09-24
cleanup: small code cleanups

Andreagit97:
While I was looking at the code, I noticed some small improvements


main.go
```diff
@@ -114,15 +114,12 @@ func main() {
 		log.Fatalf("Failed to get function addrs: %s", err)
 	}
 
-	var opts ebpf.CollectionOptions
```
Andreagit97, line 117: just moved it below nearer its usage

PR #604, opened 2025-09-25
Add riscv64 support

hack3ric:
vmlinux.h is generated on Linux 6.4; for kprobe_pwru, applied the same technique for detecting trampoline FP as on arm64.

Tested on real riscv64 hardware (Milk-V Pioneer) with Arch Linux RISC-V and Linux 6.16.


brb:
Thanks! Did you consider adding `riscv64` to the build process? Grep for `GOARCH` in `Makefile`.

PR #623, opened 2025-11-13
Fix path for available_filter_functions access

Seas0:
Updated the path for accessing available filter functions to comply with Linux v6.17 changes. Added error handling for both new and old paths.


gentoo-root (approved):

brb:
Thanks!

PR #632, opened 2025-12-30
bpf: Use get_addr() in BPF map kprobes

brb:
Previously, the addr, which was set PT_REGS_IP(ctx), returned $FN_ADDR+1. This made the addr to symbol name resolution to fail.


jschwinger233 (approved):
Happy new year~

PR #651, opened 2026-02-03
Use Cilium lumberjack/v2 fork

tklauser:
This fork fixes log compression by using a temp file and then atomically renaming to the final destination.

Ref. https://github.com/cilium/cilium/pull/19152


gyutaeb:
@tklauser Thank you!

PR #656, opened 2026-02-20
output: include date and timezone in absolute timestamp

gyutaeb:
## Summary

- Change `--timestamp absolute` format from time-only (`15:04:05.000`) to ISO 8601 date-time with milliseconds (`2006-01-02T15:04:05.000`)
- Include date for better correlation with external logs and long-running traces
- Add unit test for absolute timestamp format validation

## Motivation

The current `--timestamp absolute` output only shows time-of-day (e.g., `14:35:42.123`), which makes it difficult to use in long-running trace sessions that span multiple days or to correlate with external log sources that include full timestamps. This change outputs the full date in ISO 8601 format (e.g., `2026-02-20T14:35:42.123`).

## Test

- [x] `go vet` passes
- [x] Unit test `TestGetAbsoluteTs` validates ISO 8601 format and parseability
- [x] All existing tests pass (`go test ./internal/...`)

### DIFF
--- a/internal/pwru/output.go
+++ b/internal/pwru/output.go
@@ -32,7 +32,7 @@ import (
 	"github.com/cilium/pwru/internal/byteorder"
 )
 
-const absoluteTS string = "15:04:05.000"
+const absoluteTS string = "2006-01-02T15:04:05.000"
 
 type WriteSyncer interface {
 	io.Writer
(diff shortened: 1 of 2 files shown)


brb (approved):
Thanks!

PR #659, opened 2026-03-05
Replace per-CPU counter maps with global atomic counters

rgo3:
PR #412 introduced a sync_fetch_and_add() helper and 3 per-CPU array counter maps as a workaround for -mcpu=v1 not supporting atomic fetch instructions. Since we can enable -mcpu=v3 (BPF ISA v3 has been available since kernel 5.1, and pwru requires >= 5.3), we can revert to the simpler approach:  __sync_fetch_and_add() on global u64 variables.
This removes the custom helper and the per-CPU maps entirely. 

Note that verifying the -mcpu=v3 requirement at compile time is currently not possible with our toolchain. clang 19 always emits atomic_fetch_add regardless of -mcpu due to llvm/llvm-project#101428, which was reverted in clang 20. With clang 20, building with -mcpu=v1 correctly produces the "Invalid usage of the XADD return value" error.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,12 @@
 ARG BUILDPLATFORM
 FROM --platform=$BUILDPLATFORM docker.io/library/golang:1.26.2 AS build
 
-ARG TARGETARCH    
-RUN gcc_pkg=$(if [ "${TARGETARCH}" = "arm64" ]; then echo "aarch64"; else echo "x86-64"; fi)  && \
+ARG TARGETARCH
+RUN gcc_pkg=$(if [ "${TARGETARCH}" = "arm64" ]; then echo "aarch64"; else echo "x86-64"; fi) && \
     apt update && \
-    apt install -y make git clang-19 llvm curl gcc flex bison gcc-${gcc_pkg}* libc6-dev-${TARGETARCH}-cross && \
-    ln -s /usr/bin/clang-19 /usr/bin/clang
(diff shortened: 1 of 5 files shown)


brb:
Thanks!

Shall we also bump from clang-19 to clang-20 in our toolchain (`grep clang-19`)?

brb:
5.10 is failing with:

```
2026/03/05 14:45:21 ERROR Failed to run pwru error="Failed to load objects: Verifier error: load program: invalid argument:\n\tfunc#0 @0\n\tfunc#1 @99\n\tfunc#2 @1816\n\tfunc#3 @1976\n\tfunc#4 @2136\n\tfunc#5 @2143\n\tfunc#6 @2150\n\tfunc#7 @2154\n\tBPF_STX uses reserved fields\n\tprocessed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0\n: program kprobe_skb_2: load program: invalid argument: BPF_STX uses reserved fields (9 line(s) omitted)"
```

rgo3:
Ugh, not quite sure what's wrong with 5.10. Generally __sync_fetch_and_and should be supported on that version. I suspect that maybe the newer compilers emit instructions that those older kernels can't handle. Will investigate more next week.

I've added a temporary commit to bump clang. Can make a separate PR for this if you like.

brb:
Hmm, in the worst case 5.10 is EOL by the end of the year :-)

PR #683, opened 2026-06-20
fixed issue# 649 Report recursion_misses for each probe on exit

Aditya-132:
Screenshot From 2026-06-20 20-24-28

### DIFF
--- a/main.go
+++ b/main.go
@@ -12,6 +12,8 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"sort"
+	"strings"
 	"syscall"
 	"time"
 
@@ -286,6 +288,7 @@ func run(flags pwru.Flags) error {
 		return fmt.Errorf("Failed to load objects: %w", err)
 	}
 	defer coll.Close()
+	defer reportRecursionMisses(coll)
 
 	for _, prog := range coll.Programs {
 		prog.VerifierLog = ""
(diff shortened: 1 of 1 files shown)


brb (changes requested):
- What does the screenshot in the PR description is meant to show?
- Did you check whether `gofmt` does not fail?

Aditya-132:
> * What does the screenshot in the PR description is meant to show?
> * Did you check whether `gofmt` does not fail?

its proof that code is working

brb:
> its proof that code is working

It show `bpftool`'s output, while the change is for pwru? Do I miss something?

Aditya-132:
image

Aditya-132:
@brb check this screenshot

brb:
Thanks. Could you squash both commits into one, and then use better git commit message description than "fixed issue ..." (e.g. see https://chris.beams.io/git-commit)? Also, let's print the new log msg only if there are any misses.

Aditya-132:
> Thanks. Could you squash both commits into one, and then use better git commit message description than "fixed issue ..." (e.g. see https://chris.beams.io/git-commit)? Also, let's print the new log msg only if there are any misses.

slog warn will never reach as i already made this condition len(misses) == 0

Aditya-132:
done

brb (approved):
Thanks!

PR #690, opened 2026-07-04
fix: honor output-tunnel in JSON output

immanuwell:
`PrintJson()` checked `OutputTuple` for `tunnel_tuple`, not `OutputTunnel`.

So 2 cases get kinda weird:
- `pwru --output-json --output-tunnel --output-tuple=false` drops `tunnel_tuple`
- `pwru --output-json --output-tuple --output-tunnel=false` still prints `tunnel_tuple`

This switches the guard to `OutputTunnel` and adds regression tests. Pretty small fix, but it keeps JSON output lined up with the flags.

Repro:
1. Run `go test ./internal/pwru -run "TestPrintJSON|TestGetAbsoluteTs" -v`
2. Before the fix, the 2 `TestPrintJSON...` cases fail
3. With this patch, they pass

### DIFF
--- a/internal/pwru/output.go
+++ b/internal/pwru/output.go
@@ -243,7 +243,7 @@ func (o *output) PrintJson(event *Event) error {
 		d.Tuple = t
 	}
 
-	if o.flags.OutputTuple {
+	if o.flags.OutputTunnel {
 		t := &jsonTuple{}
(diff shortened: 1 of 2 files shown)


smagnani96 (approved):
Looks good, thanks for the catch 👍🏼

PR #691, opened 2026-07-07
fix: honor output-tcp-flags in JSON output

immanuwell:
`--output-tcp-flags` works in text output, but JSON was doing its own thing and always emitted `tuple.flags`.

Repro
`go test ./internal/pwru -run TestPrintJSONHonorsOutputTCPFlags -count=1`
Before this patch it fails, `tuple.flags` is still there when the flag is off.

Fix
Only set `tuple.flags` for JSON when `--output-tcp-flags` is enabled.
Added tests for both off and on cases.

### DIFF
--- a/internal/pwru/output.go
+++ b/internal/pwru/output.go
@@ -239,7 +239,9 @@ func (o *output) PrintJson(event *Event) error {
 		t.Sport = byteorder.NetworkToHost16(event.Tuple.Sport)
 		t.Dport = byteorder.NetworkToHost16(event.Tuple.Dport)
 		t.Proto = event.Tuple.L4Proto
-		t.Flags = event.Tuple.TCPFlag.String()
+		if o.flags.OutputTCPFlags {
+			t.Flags = event.Tuple.TCPFlag.String()
+		}
 		d.Tuple = t
 	}
 
@@ -250,6 +252,9 @@ func (o *output) PrintJson(event *Event) error {
 		t.Sport = byteorder.NetworkToHost16(event.TunnelTuple.Sport)
 		t.Dport = byteorder.NetworkToHost16(event.TunnelTuple.Dport)
 		t.Proto = event.TunnelTuple.L4Proto
+		if o.flags.OutputTCPFlags {
+			t.Flags = event.TunnelTuple.TCPFlag.String()
+		}
 		d.TunnelTuple = t
 	}
 
--- a/internal/pwru/output_test.go
+++ b/internal/pwru/output_test.go
@@ -7,7 +7,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"regexp"
-	"syscall"
 	"testing"
 	"time"
 )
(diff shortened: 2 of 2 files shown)


internal/pwru/output.go
rgo3, line 254: While we're at it, we could make sure that the tunnel json also prints and respects setting tcp flags. The regular text output does IIRC.

internal/pwru/output_test.go
```diff
@@ -92,3 +92,49 @@ func TestPrintJSONTunnelTupleOff(t *testing.T) {
 		t.Fatalf("unexpected tunnel_tuple field in json output: %s", outBuf.String())
 	}
 }
+
+func TestPrintJSONHonorsOutputTCPFlags(t *testing.T) {
```
rgo3: Could you adjust the naming scheme to the previous tests that you added? 

```suggestion
func TestPrintJSONTCPFlagsOff(t *testing.T) {
```

rgo3 (changes requested):
Thanks for the PR. Two suggestions inline.

internal/pwru/output.go
immanuwell: done

internal/pwru/output_test.go
immanuwell: done

immanuwell:
@rgo3 addressed all suggestions, PTAL

internal/pwru/output_test.go
```diff
@@ -92,3 +92,105 @@ func TestPrintJSONTunnelTupleOff(t *testing.T) {
 		t.Fatalf("unexpected tunnel_tuple field in json output: %s", outBuf.String())
 	}
 }
+
```
rgo3: All the tests in this test file except `func TestGetAbsoluteTs` do something very similar and add the same boilerplate. I think it would make sense to derive a table driven test here to make this a bit leaner. E.g. something similar to this:

```golang
func TestPrintJSONTupleFields(t *testing.T) {
	tests := []struct {
		name             string
		outputTuple      bool
		outputTunnel     bool
		outputTCPFlags   bool
		wantTuple        bool
		wantTunnelTuple  bool
		tupleField       string
		wantFlags        string
		wantFlagsPresent bool
	}{
		{
			name:           "tuple flags off",
			outputTuple:    true,
			outputTCPFlags: false,
			wantTuple:      true,
			tupleField:     "tuple",
		},
		{
			name:             "tuple flags on",
			outputTuple:      true,
			outputTCPFlags:   true,
			wantTuple:        true,
			tupleField:       "tuple",
			wantFlags:        "SYN|ACK",
			wantFlagsPresent: true,
		},
		{
			name:            "tunnel tuple flags off",
			outputTunnel:    true,
			outputTCPFlags:  false,
			wantTunnelTuple: true,
			tupleField:      "tunnel_tuple",
		},
		{
			name:             "tunnel tuple flags on",
			outputTunnel:     true,
			outputTCPFlags:   true,
			wantTunnelTuple:  true,
			tupleField:       "tunnel_tuple",
			wantFlags:        "SYN|ACK",
			wantFlagsPresent: true,
		},
	}

	for _, tt := range tests {
        ....
    }
}
```

immanuwell: @rgo3 done, refactored the repeated tests

immanuwell:
@rgo3 please take a look

rgo3 (approved):
Simplified it a bit, guess I kinda led you a stray with that many test options. Thanks for contributing!

PR #697, opened 2026-07-14
fix: preserve shared info in JSON output

immanuwell:
it fixes overwrite in JSON output

When `--output-skb` and `--output-skb-shared-info` are both enabled, both payloads were written to `skb_metadata`. The second write wiped the first, and `skb_shared_info` never showed up

Keep skb data in `skb_metadata` and write shared info to the existing `skb_shared_info` field. Both fields show up now

Repro on the unfixed build:

```sh
sudo ./pwru --filter-func=ip_rcv --output-json \
  --output-skb --output-skb-shared-info --output-limit-lines=1
```

Wait for one packet, the old output has `skb_metadata` containing the shared-info dump, but no `skb_shared_info` field

Tests: `make test`

### DIFF
--- a/internal/pwru/output.go
+++ b/internal/pwru/output.go
@@ -289,13 +289,16 @@ func (o *output) PrintJson(event *Event) error {
 		d.Stack = getStackData(event, o)
 	}
 
+	var skbData, shinfoData string
 	if o.flags.OutputSkb {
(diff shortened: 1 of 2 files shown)


brb (approved):
Thanks!

PR #699, opened 2026-07-16
fix: preserve relative timestamps in JSON output

immanuwell:
Fix relative JSON timestamps for repeated skb events. 
The cache now updates after calculating `time`, so JSON matches text output

Repro:
`sudo pwru --output-json --timestamp=relative 'icmp and host 127.0.0.1'`
Run `ping -c1 127.0.0.1`. 
Before this, later events for one `skb` show `time: 0`; now they show the delta


brb (approved):
Thanks!

PR #703, opened 2026-07-22
fix: close namespace FDs while collecting interfaces

immanuwell:
`getIfaces` kept namespace FDs open while building the default metadata cache. They pile up fast on busy nodes

Close them after the inode lookup and close the temporary namespace handles. No behavior change besides releasing them

Repro: apply this test to the parent commit, then run `go test ./internal/pwru -run TestGetIfacesDoesNotLeakNamespaceFDs -count=1`
It fails with leaked FDs. This PR passes


brb (approved):
Thanks!

PR #705, opened 2026-07-23
fix: filter packets with a zero mark

immanuwell:
`--filter-mark=0` skipped its BPF check, so a packet marked `42` still showed up. 
fix: use the mask as the filter guard.

Repro:
```sh
sudo pwru --filter-func=ip_output --filter-mark=0 --output-json --output-limit-lines=1 "udp and port 54331"
# send a UDP packet to port 54331 with SO_MARK=42
```

Before, output contains `"mark":42`. After this change, it prints no event. 
and zero marks now work too

### DIFF
--- a/bpf/kprobe_pwru.c
+++ b/bpf/kprobe_pwru.c
@@ -272,7 +272,7 @@ filter_meta(struct sk_buff *skb) {
 	if (cfg->netns && get_netns(skb) != cfg->netns) {
 			return false;
 	}
-	if (cfg->mark && cfg->mask && (BPF_CORE_READ(skb, mark) & cfg->mask) != cfg->mark) {
+	if (cfg->mask && (BPF_CORE_READ(skb, mark) & cfg->mask) != cfg->mark) {
 		return false;
 	}
 	if (cfg->ifindex != 0 && BPF_CORE_READ(skb, dev, ifindex) != cfg->ifindex) {
--- a/internal/pwru/types.go
+++ b/internal/pwru/types.go
@@ -90,7 +90,7 @@ func (f *Flags) SetFlags() {
 	flag.StringVar(&f.FilterFunc, "filter-func", "", "filter kernel functions to be probed by name (exact match, supports RE2 regular expression)")
 	flag.StringSliceVar(&f.FilterNonSkbFuncs, "filter-non-skb-funcs", nil, "filter non-skb kernel functions to be probed (--filter-track-skb-by-stackid will be enabled)")
 	flag.StringVar(&f.FilterNetns, "filter-netns", "", "filter netns (\"/proc/<pid>/ns/net\", \"inode:<inode>\")")
-	flag.Var(newMarkFlagValue(&f.FilterMark, &f.FilterMarkMask), "filter-mark", "filter skb mark (format: mark[/mask], e.g., 0xa00/0xf00)")
+	flag.Var(newMarkFlagValue(&f.FilterMark, &f.FilterMarkMask), "filter-mark", "filter skb mark (format: mark[/mask], e.g., 0xa00/0xf00; default: 0/0; exact match if mask is omitted)")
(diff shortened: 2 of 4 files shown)


bpf/kprobe_pwru.c
```diff
@@ -272,7 +272,7 @@ filter_meta(struct sk_buff *skb) {
 	if (cfg->netns && get_netns(skb) != cfg->netns) {
 			return false;
 	}
-	if (cfg->mark && cfg->mask && (BPF_CORE_READ(skb, mark) & cfg->mask) != cfg->mark) {
+	if (cfg->mask && (BPF_CORE_READ(skb, mark) & cfg->mask) != cfg->mark) {
```
gentoo-root, line 275: > --filter-mark=0 skipped its BPF check, so a packet marked 42 still showed up.

According to `pwru --help`, it looks expected:

```
--filter-mark mark[/mask]        filter skb mark (format: mark[/mask], e.g., 0xa00/0xf00) (default 0x0)
```

If 0 is the default, it makes sense that `--filter-mark=0` doesn't change the behavior.

But I agree that it is a gap in functionality, and considering that the [default mask is 0xffffffff](https://github.com/cilium/pwru/blob/bdc2eac3e503f965adffef6d076dc3d29ffa95c4/internal/pwru/types.go#L253-L254), it makes sense that `--filter-mark=0` should behave as "match 0/0xffffffff".

So, let's proceed with this change, but let's also document it properly in `--help`:
1. Say that the default is `0/0`.
2. Mention that this filter makes the exact match if the mask is omitted (i.e. the default mask is 0xffffffff). Let's make it clear and concise.

gentoo-root: I also suggest ANDing `cfg->mark` itself with `cfg->mask`, so that we only compared those bits that are part of the mask. It can be done in Go code, while parsing the config. Or, alternatively, like this:

```
!((BPF_CORE_READ(skb, mark) ^ cfg->mask) & cfg->mask)
```

gentoo-root (changes requested):

bpf/kprobe_pwru.c
immanuwell: done, addressed both

immanuwell:
@gentoo-root addressed both your suggestions, please take a look

immanuwell:
@gentoo-root PTAL

internal/pwru/types.go
```diff
@@ -90,7 +90,7 @@ func (f *Flags) SetFlags() {
 	flag.StringVar(&f.FilterFunc, "filter-func", "", "filter kernel functions to be probed by name (exact match, supports RE2 regular expression)")
 	flag.StringSliceVar(&f.FilterNonSkbFuncs, "filter-non-skb-funcs", nil, "filter non-skb kernel functions to be probed (--filter-track-skb-by-stackid will be enabled)")
 	flag.StringVar(&f.FilterNetns, "filter-netns", "", "filter netns (\"/proc/<pid>/ns/net\", \"inode:<inode>\")")
-	flag.Var(newMarkFlagValue(&f.FilterMark, &f.FilterMarkMask), "filter-mark", "filter skb mark (format: mark[/mask], e.g., 0xa00/0xf00)")
+	flag.Var(newMarkFlagValue(&f.FilterMark, &f.FilterMarkMask), "filter-mark", "filter skb mark (format: mark[/mask]; default: 0/0; exact match if mask is omitted, e.g., 0xa00/0xf00)")
```
gentoo-root: Nit: it would be clearer in this order:

```
(format: mark[/mask], e.g., 0xa00/0xf00; default: 0/0; exact match if mask is omitted)
```

gentoo-root (approved):
Sorry for the delay; the code looks good to me, one tiny comment on the wording.

internal/pwru/types.go
immanuwell: done

immanuwell:
@gentoo-root done, addressed your suggestion

gentoo-root (approved):

PR #718, opened 2026-08-07
fix: prevent output-bpfmap panics for oversized fields

### LINKED ISSUE #709: output-bpfmap: prevent slice-bounds panic for map fields larger than 256 bytes
### Summary

`printBpfmapValue.String()` trusts `KeySize` and `ValueSize` when slicing fixed `[256]byte` arrays. Either size exceeding 256 causes a Go runtime slice-bounds panic while formatting `--output-bpfmap` data.

### Affected revision

`bdc2eac3e503f965adffef6d076dc3d29ffa95c4`

### Code path

1. `printBpfmapValue` stores key and value data in fixed 256-byte arrays.
2. The structure carries independent `uint32` key/value sizes.
3. `String()` evaluates `i.Key[:i.KeySize]` and `i.Value[:i.ValueSize]` without validation.
4. Go panics when either bound exceeds 256.

Source: [`internal/pwru/types.go` lines 285-298](https://github.com/cilium/pwru/blob/bdc2eac3e503f965adffef6d076dc3d29ffa95c4/internal/pwru/types.go#L285-L298)

### Minimal reproduction

Add this test in package `pwru`:

```go
func TestPrintBpfmapValueOversize(t *testing.T) {
    v := printBpfmapValue{KeySize: 257, ValueSize: 257}
    _ = v.String()
}
```

The test panics with a slice-bounds error. The CLI path is reachable when `--output-bpfmap` formats helper data for a map whose reported key or value size exceeds 256 bytes.

### Expected behavior

pwru should truncate oversized captured data, reject an unsupported size, or report that only 256 bytes were captured.

### Actual behavior and impact

The output process terminates with a runtime panic, ending the tracing session and potentially losing diagnostic output.

### Suggested remediation

Validate both sizes before slicing. Clamp to the fixed capacity and mark the output as truncated, or return an explicit formatting error.

### Regression tests

Cover sizes 0, 256, 257, and the maximum `uint32` value for both fields. None should panic.

immanuwell:
Fixes #709

`--output-bpfmap` panics when a traced map key or value is larger than 256 bytes. Keep the real size in output, dump the captured 256-byte prefix

Repro: run pwru with `--output-bpfmap` while a TC BPF program updates a map with a 257-byte key or value. Before this it panics with `slice bounds out of range [:257] with length 256`.

### DIFF
--- a/internal/pwru/types_test.go
+++ b/internal/pwru/types_test.go
@@ -3,7 +3,13 @@
 
 package pwru
 
-import "testing"
+import (
+	"encoding/hex"
+	"fmt"
+	"math"
+	"strings"
+	"testing"
+)
 
 func TestMarkFlagValue(t *testing.T) {
 	var mark, mask uint32
@@ -25,5 +31,38 @@ func TestMarkFlagValue(t *testing.T) {
 	}
 	if mark != 0xaf || mask != 0x0f {
 		t.Fatalf("Set() = mark %#x, mask %#x; want mark %#x, mask %#x", mark, mask, uint32(0xaf), uint32(0x0f))
+	}
+}
+
+func TestPrintBpfmapValueStringTruncatesOversizedFields(t *testing.T) {
+	tests := []struct {
+		name    string
+		size    uint32
+		wantLen int
+	}{
+		{"zero", 0, 0},
+		{"maximum captured", 256, 256},
+		{"oversized", 257, 256},
+		{"maximum uint32", math.MaxUint32, 256},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			v := printBpfmapValue{KeySize: tt.size, ValueSize: tt.size}
+			for i := range v.Key {
+				v.Key[i] = byte(i)
+				v.Value[i] = byte(i)
+			}
+
+			got := v.String()
+			want := fmt.Sprintf(
+				"key(%d):\n%svalue(%d):\n%s",
+				tt.size, hex.Dump(v.Key[:tt.wantLen]),
+				tt.size, hex.Dump(v.Value[:tt.wantLen]),
+			)
+			if !strings.Contains(got, want) {
+				t.Fatalf("String() missing expected dump:\n%s", got)
+			}
+		})
 	}
 }
--- a/internal/pwru/types.go
+++ b/internal/pwru/types.go
@@ -289,8 +289,8 @@ type printBpfmapValue struct {
 }
 
 func (i *printBpfmapValue) String() string {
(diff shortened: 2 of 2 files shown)


internal/pwru/types_test.go
```diff
@@ -27,3 +33,39 @@ func TestMarkFlagValue(t *testing.T) {
...
+	}{
+		{"zero key", 0, 1, 0, 1},
+		{"maximum captured key", 256, 1, 256, 1},
+		{"oversized key", 257, 1, 256, 1},
+		{"maximum uint32 key", math.MaxUint32, 1, 256, 1},
+		{"zero value", 1, 0, 1, 0},
+		{"maximum captured value", 1, 256, 1, 256},
+		{"oversized value", 1, 257, 1, 256},
+		{"maximum uint32 value", 1, math.MaxUint32, 1, 256},
+	}
```
rgo3, line 47: Nit: Given that it's the exact same code I think we could squash some test cases and don't need to test key and value independently:

```suggestion
	tests := []struct {
		name      string
		size   uint32
		wantLen    int
	}{
		{"zero key", 0, 0},
		{"maximum captured", 256, 256},
		{"oversized", 257, 256},
		{"maximum uint32 key", math.MaxUint32, 256},
	}
```

internal/pwru/types_test.go
```diff
@@ -27,3 +33,39 @@ func TestMarkFlagValue(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			v := printBpfmapValue{KeySize: tt.keySize, ValueSize: tt.valueSize}
+			for i := range v.Key {
+				v.Key[i] = byte(i)
+				v.Value[i] = byte(i)
+			}
+
+			got := v.String()
+			want := "key(" + strconv.FormatUint(uint64(tt.keySize), 10) + "):\n" + hex.Dump(v.Key[:tt.keyLen]) +
+				"value(" + strconv.FormatUint(uint64(tt.valueSize), 10) + "):\n" + hex.Dump(v.Value[:tt.valueLen])
```
rgo3:
```suggestion
			want := fmt.Sprintf(
                                "key(%d):\n%svalue(%d):\n%s",
                                tt.keySize, hex.Dump(v.Key[:tt.keyLen]),
                                tt.valueSize, hex.Dump(v.Value[:tt.valueLen]),
                        )
```

Not 100% percent sure if this works, but it feels more readable to me.

rgo3 (approved):
Overall LGTM, two small nits for easier readabiliyt 🙏

immanuwell:
@rgo3 applied your suggestions

PR #724, opened 2026-08-21
libpcap: bump vendored copy to 1.10.6

tklauser:
Downloaded from https://github.com/the-tcpdump-group/libpcap/archive/refs/tags/libpcap-1.10.6.zip

No changes necessary anymore, the `.gitignore` entry for `bpf_filter.c` is gone upstream.


brb (approved):
Thanks!

PR #725, opened 2026-08-22
fix: propagate plain-text output errors

### LINKED ISSUE #711: output: propagate plain-text writer failures
### Summary

The plain-text output path ignores both return values from `io.Writer.Write`. Disk-full, broken-pipe, closed-file, or other sink failures do not propagate to the capture loop, and pwru may continue tracing after output has stopped.

### Affected revision

`bdc2eac3e503f965adffef6d076dc3d29ffa95c4`

### Code path

1. `Print()` builds the text record.
2. It calls `o.writer.Write(...)`.
3. The returned count and error are discarded.
4. `Print()` has no error return, so the caller cannot stop or report failure.

Source: [`internal/pwru/output.go` lines 591-593](https://github.com/cilium/pwru/blob/bdc2eac3e503f965adffef6d076dc3d29ffa95c4/internal/pwru/output.go#L591-L593)

The JSON path already returns encoding errors: [`internal/pwru/output.go` lines 302-310](https://github.com/cilium/pwru/blob/bdc2eac3e503f965adffef6d076dc3d29ffa95c4/internal/pwru/output.go#L302-L310).

### Minimal reproduction

```go
type failingWriter struct{}

func (failingWriter) Write(p []byte) (int, error) {
    return 0, errors.New("sink failed")
}
```

Configure `output` with this writer and call `Print()`. The function returns normally and the error is lost.

### Expected behavior

A write failure should propagate to the event loop, be logged once with context, and terminate or explicitly degrade the capture.

### Actual behavior and impact

The error is silently ignored. Users can believe a trace was saved successfully even after the output sink stopped accepting data.

### Suggested remediation

Change `Print()` to return an error, check for short writes, and handle failure consistently with `PrintJSON()`.

### Regression tests

Verify propagation of a normal writer error and an `io.ErrShortWrite` condition.

immanuwell:
Plain-text output ignored writer errors, so pwru could exit 0 after losing trace data

This change returns write and short-write errors to the capture loop

Repro:

```sh
sudo ./pwru --backend=kprobe --filter-func=ip_rcv \
  --output-file=/dev/full --output-limit-lines=1 'icmp' &
pid=$!
sleep 2
ping -c1 127.0.0.1
wait "$pid"
```

Before this fix the write failed but pwru exited 0. 
Now it reports ENOSPC and exits 1

Fixes: #711

### DIFF
--- a/internal/pwru/output.go
+++ b/internal/pwru/output.go
@@ -497,7 +497,7 @@ func fprintWithPadding(writer *os.File, data string, maxLenSeen *int) {
 	fmt.Fprintf(writer, formatter, data)
 }
 
-func (o *output) Print(event *Event) {
+func (o *output) Print(event *Event) error {
 	var sb strings.Builder
 	sb.Grow(256)
 
@@ -589,7 +589,15 @@ func (o *output) Print(event *Event) {
 	}
 
 	sb.WriteString("\n")
-	o.writer.Write([]byte(sb.String()))
+	line := []byte(sb.String())
+	n, err := o.writer.Write(line)
+	if err != nil {
+		return err
+	}
+	if n != len(line) {
+		return io.ErrShortWrite
+	}
+	return nil
 }
 
 func (o *output) getIfaceName(netnsInode, ifindex uint32) string {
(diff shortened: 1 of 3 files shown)


gentoo-root (approved):
Thanks for the fix!

PR #729, opened 2026-08-30
feat: add distro packaging docs and nfpm config

### LINKED ISSUE #89: Package pwru for major distros
Each pwru release published in https://github.com/cilium/pwru/releases is statically linked, and it doesn't have any runtime dependencies. So, it should be trivial to package for major distros (Archlinux, Ubuntu, Debian, Fedora, etc).

amarkdotdev:
## Summary
- Add nfpm config and packaging guide for Debian, RPM, and Arch

Fixes #89

## Test plan
- [x] Verified nfpm.yaml syntax
- [x] Reviewed packaging.md instructions

### DIFF
--- a/docs/packaging.md
+++ b/docs/packaging.md
@@ -1,0 +1,38 @@
+# Packaging pwru for Linux distributions
+
+pwru release binaries are statically linked and have no runtime dependencies beyond a supported kernel. This makes packaging straightforward for major distributions.
+
+## Build a release binary
+
+```shell
+make release
+```
+
+Artifacts are written to `release/`.
+
+## Debian and Ubuntu (.deb)
+
+Use [nfpm](https://nfpm.goreleaser.com/) with the provided config:
+
+```shell
+export VERSION=1.0.0
+export ARCH=amd64
+nfpm package --config packaging/nfpm.yaml --packager deb
+```
+
+## Fedora and RHEL (.rpm)
+
+```shell
+export VERSION=1.0.0
+export ARCH=amd64
+nfpm package --config packaging/nfpm.yaml --packager rpm
+```
+
+## Arch Linux
+
+See the [Arch Linux package](https://archlinux.org/packages/extra/x86_64/pwru/) for a maintained source build.
+
+## Notes
+
+- pwru requires Linux kernel >= 5.3 (see README for full kernel config requirements).
+- Release binaries are published on each GitHub release. Downstream packages should prefer building from source when packaging guidelines require it.
--- a/packaging/nfpm.yaml
+++ b/packaging/nfpm.yaml
@@ -1,0 +1,17 @@
+name: pwru
+arch: "${ARCH}"
+platform: linux
+version: "${VERSION}"
+section: utils
+priority: optional
+maintainer: Cilium Authors <cilium@linuxfoundation.org>
+description: eBPF-based tool for tracing network packets in the Linux kernel
+homepage: https://github.com/cilium/pwru
+license: Apache-2.0
+contents:
+  - src: release/pwru-linux-${ARCH}
+    dst: /usr/bin/pwru
+    file_info:
+      mode: 0755
+  - src: README.md
+    dst: /usr/share/doc/pwru/README.md


docs/packaging.md
```diff
@@ -0,0 +1,47 @@
...
+## Arch Linux (PKGBUILD sketch)
+
+```shell
+pkgname=pwru
+pkgver=1.0.0
+pkgrel=1
+arch=('x86_64' 'aarch64')
+source=()
+package() {
+  install -Dm755 release/pwru-linux-${CARCH} "${pkgdir}/usr/bin/pwru"
```
gentoo-root: As far as I understand Arch Linux's [guidelines](https://wiki.archlinux.org/title/AUR_submission_guidelines), packages must be built from the source code, whenever it's available (i.e. the software is not proprietary):

> Packages that use prebuilt deliverables, when the sources are available, must use the -bin suffix. An exception to this is with Java.

AUR packages can be prebuilt, which makes little sense for such a tiny tool like pwru, but even considering this pkgname would have to be pwru-bin, not pwru.

I suggest writing a proper PKGBUILD that builds the package from source instead.

Besides, @Foxboron [already packaged](https://github.com/cilium/pwru/issues/89#issuecomment-1497417178) it for Arch in 2023 and wrote a [proper PKGBUILD](https://github.com/Foxboron/archlinux-pkgbuilds/blob/master/pwru/PKGBUILD). The version in the repository is current (1.0.12), so we don't seem to need one more Arch package anymore.

Can't speak for Debian and Fedora packaging rules; inviting @brb (https://github.com/cilium/pwru/issues/89) for an extra pair of eyes.

gentoo-root (changes requested):

docs/packaging.md
Foxboron: The packaging here seems misguided and doesn't solve a problem upstream has?

amarkdotdev: swapped the Arch sketch for a link to Foxboron's source PKGBUILD

docs/packaging.md
```diff
@@ -28,20 +28,11 @@ export ARCH=amd64
 nfpm package --config packaging/nfpm.yaml --packager rpm
 ```
 
-## Arch Linux (PKGBUILD sketch)
+## Arch Linux
 
-```shell
```
gentoo-root: It should be one commit. You are adding this chunk in the first commit and then immediately removing it in the next one.

amarkdotdev:
updated Arch section to point at Foxboron's source PKGBUILD in 0a862ff

docs/packaging.md
```diff
@@ -28,20 +28,11 @@ export ARCH=amd64
...
-pkgname=pwru
-pkgver=1.0.0
-pkgrel=1
-arch=('x86_64' 'aarch64')
-source=()
-package() {
-  install -Dm755 release/pwru-linux-${CARCH} "${pkgdir}/usr/bin/pwru"
-}
-```
+Arch packages must be built from source when upstream sources are available. See the maintained [Arch PKGBUILD](https://github.com/Foxboron/archlinux-pkgbuilds/blob/master/pwru/PKGBUILD) for an example that builds pwru from this repository.
```
gentoo-root: If adding a link, I'd better put https://archlinux.org/packages/extra/x86_64/pwru/, it's more useful and points to an up-to-date PKGBUILD.

No need to recap our PR discussion in the docs. Is it LLM-generated?

I'm not sure it solves https://github.com/cilium/pwru/issues/89 — if I got the idea correctly, the outcome should be packages in various distro's repositories, not just instructions how to build them, which I'm not sure follow these distro's best practices (like building from sources). Could you please explain what the intended purpose and value of this contribution is?

gentoo-root (changes requested):

docs/packaging.md
Foxboron: I resigned from Arch Linux a month ago. The link that is being used here won't be updated. Please use the `archlinux.org` link as mentioned above.

docs/packaging.md
amarkdotdev: squashed into one commit

docs/packaging.md
amarkdotdev: switched to the archlinux.org link and dropped the extra commentary

gentoo-root:
As mentioned in https://github.com/cilium/pwru/pull/729#discussion_r3895625431, I don't understand the value and purpose of this pull request, and I didn't get a response to my question, so at this point I'm stopping reviewing this PR, unless it evolves into something meaningful.

My understanding of https://github.com/cilium/pwru/issues/89 is that we want packages in various Linux distributions. Adding docs on DIY package creation doesn't resolve this issue. I also see that our [README](https://github.com/cilium/pwru/blob/ad26c347949bb69a5177ef44d810f9a2c98eb53d/README.md#distro-packages) already has a table that tracks packaging status in various distros, as well as the `make release` build instructions, so your new [docs/packaging.md](https://github.com/cilium/pwru/pull/729/changes/46685559638de1e020bb17df1ea93bb6acf10a9f#diff-e04d95ab6f602fbd2a4be399b2d917253b3b494a45c8d4d5c07326d06710b60f) doesn't really add anything. The users should read the existing README, Arch Linux users will see that there is an up-to-date version in the repository, Debian and Ubuntu users will see that there is an older version, and they'll decide whether to install that or build manually, Fedora users will have to build manually for now, and the existing table includes many more distributions.

What could be helpful in regard to https://github.com/cilium/pwru/issues/89 is working towards including pwru into Fedora repositories, and updating the version in Debian-based distributions. If you'd like to make a useful contribution, this direction would be helpful. Thanks for understanding!

docs/packaging.md
amarkdotdev: still a single commit on head (4668555), the squash dropped the churn

docs/packaging.md
amarkdotdev: re #89: mostly nfpm docs + sample config so folks can build debs/rpms without guessing. not trying to replace official distro packages. fine to close if maintainers dont want it

amarkdotdev:
@gentoo-root squash + archlinux.org link are on head, resolved the threads. lmk if the nfpm docs still make sense for you

brb:
As @gentoo-root pointed in [the comment above](https://github.com/cilium/pwru/pull/729#issuecomment-5525625933), #89 is about getting pwru included in distro package managers. We don't want to be responsible for the packaging software.

PR #732, opened 2026-09-04
fix: preserve CPU 0 in JSON output

immanuwell:
JSON events generated on CPU 0 do not contain a cpu field.
it happens because Cpu uses the `json:"cpu,omitempty"` tag, and zero is a valid CPU ID

The fix is to remove omitempty so CPU 0 is serialized as `"cpu":0`
Add a regression test that serializes an event on CPU 0 and verifies the key is present

Repro:

run pwru with JSON output and pin loopback traffic to CPU 0:
```sh
sudo ./pwru --output-json --output-limit-lines=20 \
  --ready-file=/tmp/pwru.ready \
  'icmp and host 127.0.0.1' >/tmp/pwru.json &

while [ ! -e /tmp/pwru.ready ]; do sleep 0.1; done
taskset -c 0 ping -c 1 127.0.0.1

jq -s '{
  events: length,
  missing_cpu: map(select(has("cpu") | not)) | length
}' /tmp/pwru.json
```

Before the fix, all 20 events lacked the cpu key.
After the fix, they contain `"cpu":0`

Output from nonzero CPUs is unchanged, `make test` passes.
No related issue was found

AI Influence Level: AIL 1

### DIFF
--- a/internal/pwru/output_test.go
+++ b/internal/pwru/output_test.go
@@ -58,116 +58,147 @@ func TestGetAbsoluteTs(t *testing.T) {
 	}
 }
 
-func TestPrintJSONTupleFields(t *testing.T) {
+func TestJSONOutput(t *testing.T) {
 	const (
 		tcpFlagSYN tcpFlag = 1 << 1
 		tcpFlagACK tcpFlag = 1 << 4
 		wantFlags          = "SYN|ACK"
 	)
 
-	tests := []struct {
-		name           string
-		outputTunnel   bool
-		outputTCPFlags bool
-	}{
-		{
-			name:           "tuple flags off",
-			outputTCPFlags: false,
-		},
-		{
-			name:           "tuple flags on",
-			outputTCPFlags: true,
-		},
-		{
-			name:           "tunnel tuple flags off",
-			outputTunnel:   true,
-			outputTCPFlags: false,
-		},
-		{
-			name:           "tunnel tuple flags on",
-			outputTunnel:   true,
-			outputTCPFlags: true,
-		},
+	tupleEvent := func() []*Event {
+		event := newBenchmarkEvent()
+		event.Tuple.TCPFlag = tcpFlagSYN | tcpFlagACK
+		event.TunnelTuple = event.Tuple
+		return []*Event{event}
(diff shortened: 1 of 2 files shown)


internal/pwru/output_test.go
```diff
@@ -135,6 +135,25 @@ func TestPrintJSONTupleFields(t *testing.T) {
 	}
 }
 
+func TestPrintJSONIncludesCPUZero(t *testing.T) {
```
tklauser: I wonder if rather than adding a new test for each JSON output issue that we discover, it would make more sense to consolidate the `Test*JSON` tests into a single, more generic `TestJSONOutput` that would cover all these cases. What do you think?

immanuwell: hmm, those tests all share the same shape, so a table-driven TestJSONOutput would fit well I think.
I can do the consolidation here if you prefer

tklauser: Yeah, would be nice to do it in a separate commit on this PR before the actual fix and then just add the necessary changes for the CPU 0 case to the table in the fix commit.

immanuwell: @tklauser done

tklauser:
To me it looks like generative AI was used to prepare parts of the change. Please add a short paragraph to the PR description that states how you used machine learning models (including LLMs and other generative AI), and indicate the rating using [AI Influence Level](https://danielmiessler.com/blog/ai-influence-level-ail). Note that the [standard pull request template](https://github.com/cilium/pwru/blob/main/.github/pull_request_template.md?plain=1) has a checkbox for this. What's the reason you didn't use the provided PR template when opening the PR?

immanuwell:
@tklauser
> To me it looks like generative AI was used to prepare parts of the change

yes, I used Claude Code to review my changes to be sure I didn't forget something and to be sure my changes really fully tested end-to-end

> What's the reason you didn't use the provided PR template when opening the PR?

I am sorry, somehow I thought it's not neccessary for minor AI assistance

so this is AIL 1, minor AI assistance (mostly checking my PR before I send it)

tklauser (approved):

